PR #2, opened 2024-10-22
Ensure 'Apply simulation state' is always stored in its own snapshot, separate from other user actions

HungryProton:
Fixes the following bug:

+ Run a simulation
+ Create an atom (or do anything not whitelisted)
+ Undo
-> The simulation AND user operation is reverted, instead of only the last operation.

This PR applies the simulation and the user action in two separate snapshots instead.

### DIFF
--- a/godot_project/project_workspace/workspace_context/history/history.gd
+++ b/godot_project/project_workspace/workspace_context/history/history.gd
@@ -45,6 +45,10 @@ const ACTION_WHITELIST_DURING_SIMULATION: Dictionary = {
 	"Reset Color Override" : true,
 }
 
+const REDUNDANT_ACTION_WHITELIST := {
+	"Apply Simulation State": true,
+}
+
 var _workspace_context: WorkspaceContext
 var _snapshotable_systems: Array[Object]
 var _snapshot_stack: Array[Dictionary]
@@ -66,7 +70,8 @@ func register_snapshotable(in_system: Object) -> void:
 	_snapshotable_systems.append(in_system)
 
 
-func create_snapshot(in_snapshot_name: String) -> void:
+func create_snapshot(in_snapshot_name: String) -> Dictionary:
+	_monitor_redundant_snapshots(in_snapshot_name)
 	var snapshot: Dictionary = {
 		#"name" : _next_snapshot_name
 	}
@@ -75,14 +80,14 @@ func create_snapshot(in_snapshot_name: String) -> void:
 		snapshot[snapshotable] = snapshotable.create_state_snapshot()
 	
 	_add_snapshot_to_stack(snapshot, in_snapshot_name)
-	if _last_snapshot_taken_at_frame == Engine.get_frames_drawn():
-		# (can be removed if we have an use case)
-		# notice external snapshots pushed by `push_snapshot(in_snapshot, in_snapshot_name)`
-		# are not considered by this check
-		push_error("Possibly redundant snapshot detected. previous: ", _last_snapshot_name,
-					" , current: ", in_snapshot_name)
-	_last_snapshot_taken_at_frame = Engine.get_frames_drawn()
-	_last_snapshot_name = in_snapshot_name
+	return snapshot
+
+
+func push_and_apply_snapshot(in_snapshot: Dictionary, in_snapshot_name: String) -> void:
+	assert(_validate_extern_snapshot(in_snapshot), "Invalid snapshot data provided")
+	_add_snapshot_to_stack(in_snapshot, in_snapshot_name)
+	_stack_pointer -= 1
+	apply_next_snapshot()
 
 
 func _add_snapshot_to_stack(in_snapshot: Dictionary, in_snapshot_name: String) -> void:
@@ -101,6 +106,8 @@ func _add_snapshot_to_stack(in_snapshot: Dictionary, in_snapshot_name: String) -
 		_stack_pointer -= 1
 	
 	_version += 1
+	_last_snapshot_taken_at_frame = Engine.get_frames_drawn()
+	_last_snapshot_name = in_snapshot_name
 	changed.emit()
 	snapshot_created.emit(in_snapshot_name)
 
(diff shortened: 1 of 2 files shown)


godot_project/project_workspace/workspace_context/history/history.gd
```diff
@@ -180,6 +187,18 @@ func _validate_extern_snapshot(in_snapshot: Dictionary) -> bool:
 	return true
 
 
+## Detects if create_snapshot is called multiple times in a single frame
+## External snapshots pushed by `push_and_apply_snapshot(in_snapshot, in_snapshot_name)`
+## are not considered by this check
+func _monitor_redundant_snapshots(in_snapshot_name: String) -> void:
+	if _last_snapshot_taken_at_frame != Engine.get_frames_drawn():
+		return
+	if REDUNDANT_ACTION_WHITELIST.has(in_snapshot_name) and in_snapshot_name != _last_snapshot_name:
```
kubecz3k, line 196: why we need the second part of this condition? `in_snapshot_name != _last_snapshot_name`  
As far as I understand what it will do it will make this error print a bit less helpful, I assume there probably should not be many situations where snapshot is created couple times in the same frame no matter the source of the snapshot (it should always react to user actions and it's uncommon for user to perform multiple actions in the same frame)

godot_project/project_workspace/workspace_context/workspace_context.gd
```diff
@@ -1202,14 +1202,33 @@ func snapshot_moment(in_operation_name: String) -> void:
...
-	# Apply simulation if the operation modified the structure of the project
-	if not History.is_operation_whitelisted(in_operation_name):
-		apply_simulation_if_running()
-	
 	# apply overdue signals, ensure snapshots contain up to date state
 	_emit_selection_in_structures_changed()
 	_emit_structure_content_changed()
 	
+	# Apply simulation if the operation modified the structure of the project
+	if not History.is_operation_whitelisted(in_operation_name) and is_simulating():
```
kubecz3k, line 1210: I think this is one of those rare cases where it's worth to consider the condition order, I would swap it. 
The key for the person who reads this code is to understand it only applies to the case where simulation is involved so I think 
`if is_simulating() and not History.is_operation_whitelisted(in_operation_name):`
will be easier to read

godot_project/project_workspace/workspace_context/workspace_context.gd
```diff
@@ -1202,14 +1202,33 @@ func snapshot_moment(in_operation_name: String) -> void:
...
+		var final_snapshot: Dictionary = _history.create_snapshot(in_operation_name)
+		var current_simulation_time: float = _simulation.get_last_seeked_time()
+		
+		# Go back just before starting the simulation
+		_history.apply_previous_snapshot()
+		
+		# Rewind simulation to the latest point and create a snapshot.
+		# This will drop the user operation (final_snapshot) from history. 
+		seek_simulation(current_simulation_time)
+		apply_simulation_if_running()
```
kubecz3k, line 1226: I believe this logic is correct, but after executing this line I would expect for NanoStructure to have exactly the same state as it was before line 1217, can you remind me why this is not the case?

kubecz3k (commented):
Applying that previous snapshot seems a bit redundant to me, why we cannot just create snapshot twice? Once just before user is performing not whitelisted operation and second time directly after that operation? That way we also would not need to add new api to History

godot_project/project_workspace/workspace_context/history/history.gd
MarianoGnu: if i understand correctly, this is because the point of this method is to catch the case where 2 snapshots **of the same type** are executed **in the same frame** so this early return is simply a whitelist of "known snapshots" that will in fact be created in the same frame 🤔

godot_project/project_workspace/workspace_context/workspace_context.gd
MarianoGnu, line 1210: unrelated to this PR, but I dont like the function name `History.is_operation_whitelisted()` it lacks contexts imho. What means to be whitelisted? it made more sense being part of SimulationStateCache because that class was about simulation, but History is an overall thing.

godot_project/project_workspace/workspace_context/workspace_context.gd
MarianoGnu: this is not the case. in line 1217 is the sate of the simulation + the change of the user (in example create an atom)
in this case is the state of the simulation without the input of the user (without that extra atom)

In the past we used `NanoUndoRedo.create_action()` to detect an event was about to happen before it happened, this is no longer the case with the snapshot system, by the time you execute `snapshot_moment` the change is already made, so all of this is to rewind that last change.

MarianoGnu (approved):

HungryProton:
> Applying that previous snapshot seems a bit redundant to me, why we cannot just create snapshot twice? Once just before user is performing not whitelisted operation and second time directly after that operation? That way we also would not need to add new api to History

The issue is when calling `snapshot_moment`, the action was already done, so the only way to take a snapshot before the user performs an action is to go back everywhere where this method is called an call apply_simulation_if_running() before doing anything.

It's possible, and removes the need for a simulation whitelist, but it's also more intrusive and error prone (I think it will be easy to forget applying the simulation in unrelated parts of the code).
The way its done here is just trying to mimic what we had before with NanoUndoRedo, but we can change that if necessary.

godot_project/project_workspace/workspace_context/history/history.gd
HungryProton: That second part is there to limit false negatives.

If we don't check that, then two accidental 'Apply simulation state' in the same frame would not be detected by this method. Even if they are in the redundant whitelist, there's no reason to have two of these actions in the same frame.

Thinking about it, this condition isn't even complete because it makes the snapshot order important. I should also check if the last_snapshot_name is whitelisted or not.

godot_project/project_workspace/workspace_context/workspace_context.gd
HungryProton: Good catch

HungryProton: Renamed it to "is_operation_whitelisted_during_simulation", not perfect but maybe we can move the list itself outside the of the History class

PR #5, opened 2024-10-22
Add attributions for used sound effects

MarianoGnu:
- A custom resource was created to collect information about third party assets
- A custom control is used to dinamically populate information about audio streams
- It is possible to enable playback of the sfx in the about dialog, but is disabled by default

-----------
Task:
- Add Open Source Information for Audio Files included in the project


brunosxs (approved):
Tested this locally, no regressions.

PR #9, opened 2024-10-22
Block UI hover detection while transform gizmo is moved

JanisMancevics:
REPRODUCTION STEPS
- Grab the transform gizmo xy movement tool and move the selection around so that mouse hovers over various UI elements
- Observe that most UI interactive controls become activated as the mouse hovers over them

------

Task: UI elements become active, while transform gizmo is grabbed


JanisMancevics:
The PR commit was created from a patch file with `git apply` and it was formed correctly.

PR #13, opened 2024-10-23
Block UI hover detection while transform gizmo is moved

JanisMancevics:
REPRODUCTION STEPS
- Grab the transform gizmo xy movement tool and move the selection around so that mouse hovers over various UI elements
-- Or any other transform gizmo handle.
- Observe that most UI interactive controls become activated as the mouse hovers over them

------

Task: UI elements become active, while transform gizmo is grabbed

### DIFF
--- a/godot_project/autoloads/ui_blocker/ui_blocker.gd
+++ b/godot_project/autoloads/ui_blocker/ui_blocker.gd
@@ -1,0 +1,24 @@
+extends CanvasLayer
+
+var _blocker: Control = null
+
+
+func _ready() -> void:
+	_blocker = %BlockerRect as Control
+	_blocker.set_mouse_filter(_blocker.MOUSE_FILTER_STOP)
+
+
+func _process(in_delta: float) -> void:
+	_unblock_last_frame_input_events()
+
+
+func is_blocking() -> bool:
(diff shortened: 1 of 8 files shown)


godot_project/autoloads/ui_blocker/ui_blocker.gd
```diff
@@ -0,0 +1,24 @@
...
+
+func _process(in_delta: float) -> void:
+	_unblock_input()
+
+
+func is_blocking() -> bool:
+	return _blocker_rect.visible
+
+
+func block_input() -> void:
```
kubecz3k, line 19: nit: when calling this function from outside I would expect it's permanent operation until I will manually unblock, maybe we could rename it to `block_current_frame_inputs` ?

kubecz3k (approved):
It's simple and I don't have any better Idea, just wondering if if the CanvasLayer node  + ColorRect is really necessary since I think I would try to do it just with plain Control utilizing mouse filter settings

godot_project/autoloads/ui_blocker/ui_blocker.gd
JanisMancevics: Done!

PR #14, opened 2024-10-24
Removed unused linking steps for libs from capnp and kj

brunosxs:
Regression on capnpSCsub during migration to github

-----------

The following file was reverted during the migration to github. This adds back the changes done on the last days before open-sourcing.


HungryProton:
I don't see these changes in the old repository, were they merged on main?

brunosxs:
Not sure then... @hpvb commented during our meeting that this file had  been lost in the process and that I should have a look at it. This is the result.

PR #18, opened 2024-10-24
Add a placeholder for Simulation Toolkit

JanisMancevics:
In the 3 settings panels for Simulation Force Fields (Relax,Molecular Simulation, and Validate), add an option for "Molecular Simulation Toolkit"

This should be a pull-down control:
The default option should be "OpenMM"
There should be a grayed out option for "LAMMPS (under development)"

So, the user won't be able to choose anything other than OpenMM. This option should be placed below the Force-Field selectors.


JanisMancevics:
This is an urgent change approved by Jonathan, as it doesn't remove any existing code or data, just adds new stuff and is a non functional features it should be safe to merge.

PR #20, opened 2024-10-25
Fix the modern, academic color switching

JanisMancevics:
The bug as written is fixed, however, the bg color varies slightly between Academic and Modern. Also, setting the colors back to Default does not set the color swatches back to the default.

- Go to workspace settings
- Switch to academic view
- Change the background color
- Switch to modern view
- Switch back to academic view
- Observe that the background color is reset to the default, but the settings still shows that the new color is being used

-------

Task: Modern view resets the user specified colors in Academic view

### DIFF
--- a/godot_project/theme/theme_3d/available_themes/flat_theme/flat_environment.tres
+++ b/godot_project/theme/theme_3d/available_themes/flat_theme/flat_environment.tres
@@ -10,11 +10,10 @@ sky_material = SubResource("PanoramaSkyMaterial_u41f5")
 
 [resource]
 background_mode = 1
-background_color = Color(0.920995, 0.920995, 0.920995, 1)
-background_energy_multiplier = 0.9
+background_color = Color(0.74978, 0.74978, 0.74978, 1)
+background_energy_multiplier = 2.0
 sky = SubResource("Sky_1u7jr")
 ambient_light_source = 2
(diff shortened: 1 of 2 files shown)


godot_project/theme/theme_3d/available_themes/flat_theme/flat_environment.tres
```diff
@@ -10,11 +10,10 @@ sky_material = SubResource("PanoramaSkyMaterial_u41f5")
 
 [resource]
 background_mode = 1
-background_color = Color(0.920995, 0.920995, 0.920995, 1)
-background_energy_multiplier = 0.9
+background_color = Color(0.74978, 0.74978, 0.74978, 1)
+background_energy_multiplier = 2.0
```
kubecz3k, line 14: is this change intentional?

kubecz3k (approved):
Approving since it's doing with the card described in ticket, but I have a bit of doubt on if it's what actually should be happening, especially that this solution forces us to have similar environment between themes which seems problematic to me

I have feeling that user defined colors should be defined per theme, ex when someone change default color in academic theme after changing to modern the modern default colors should be in use (unless user changes this settings again for modern theme), and then if one goes back to academic view the custom colors from academic view should be re-estabilished

PR #23, opened 2024-10-30
Fix a case where hovering a group in sticks mode highlights only the bond and not the entire group

MarianoGnu:
-----------
 Task: BUG: When hovering a group in **Sticks** or **Enhanced Sticks** representation, all the bonds of the group doesn't highlight, only the bond under the cursor
 
 Preview:
image


brunosxs (approved):
I see no issues, lgtm

PR #25, opened 2024-10-30
Push WIP changes to Workspace even before creating an Snapshot when done editing

MarianoGnu:
This allows to Save changes with Ctrl+S without losing them while the UI is focused

-----------
 Task: BUG: Project Description and authors are not applyed until text edit edit is unfocused


brunosxs (approved):
LGTM

PR #26, opened 2024-10-31
Fix a corner case when growing a selection left some bonds unselected

MarianoGnu:
-----------
 Task: Growing this selection does not select the pointed bond

Before:
image

After:
image

### DIFF
--- a/godot_project/project_workspace/workspace_context/structure_context/selection_db/atom_selection/atom_selection.gd
+++ b/godot_project/project_workspace/workspace_context/structure_context/selection_db/atom_selection/atom_selection.gd
@@ -233,7 +233,12 @@ func can_grow_selection() -> bool:
 
 
 func grow_selection() -> AtomSelectionResult:
-	var atoms_to_select: PackedInt32Array = PackedInt32Array()
+	var atoms_to_select_dict: Dictionary = {}
+	# In the list of atoms to select we include already selected atoms,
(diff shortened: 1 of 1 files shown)


HungryProton (approved):

godot_project/project_workspace/workspace_context/structure_context/selection_db/atom_selection/atom_selection.gd
```diff
@@ -233,7 +233,12 @@ func can_grow_selection() -> bool:
 
 
 func grow_selection() -> AtomSelectionResult:
-	var atoms_to_select: PackedInt32Array = PackedInt32Array()
+	var atoms_to_select_dict: Dictionary = {}
+	# In the list of atoms to select we include already selected atoms,
+	# this helps to automatically expand selections of bonds in some corner cases
+	for atom_id: int in get_atoms_selection():
+		atoms_to_select_dict[atom_id] = true
+
```
kubecz3k, line 241: I'm wondering if it would not be more straightforward if we would replace logic of this function with something like (pseudocode):
```
var atom_selection = get_atoms_selection()
for atom_id in atom_selection:
   var related_bonds = atom_get_bonds(atom_id)
   for bond_id in related_bonds:
      (...)
```

kubecz3k (approved):

PR #33, opened 2024-10-31
Fix a corner case not allowing to use the TransformGizmo…

MarianoGnu:
… when it overlapped the Workspace Tabs transparent space

-----------
 Task: I can place the [←↕→] Gizmo in a position where is visible but not interactible

Before:
https://github.com/user-attachments/assets/440cf1a5-861f-483e-9096-e3f09235bdb7

After:
https://github.com/user-attachments/assets/d49f781e-6653-405b-bc51-3a47d0d8aea0

### DIFF
--- a/godot_project/editor/controls/workspace_view/tab_bar_active_workspaces.gd
+++ b/godot_project/editor/controls/workspace_view/tab_bar_active_workspaces.gd
@@ -40,6 +40,22 @@ func _ready() -> void:
 	MolecularEditorContext.homepage_activated.connect(_on_homepage_activated)
 
 
+func _process(delta: float) -> void:
+	var tab_bar_rect: Rect2 = tab_bar_active_workspaces.get_global_rect()
+	var mouse_pos: Vector2 = get_global_mouse_position()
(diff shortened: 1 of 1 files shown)


godot_project/editor/controls/workspace_view/tab_bar_active_workspaces.gd
```diff
@@ -40,6 +40,22 @@ func _ready() -> void:
 	MolecularEditorContext.homepage_activated.connect(_on_homepage_activated)
 
 
+func _process(delta: float) -> void:
+	var tab_bar_rect: Rect2 = tab_bar_active_workspaces.get_global_rect()
```
kubecz3k, line 44: do we need to perform this in process? Is there any 'mouse event' signal that we could use instead?

kubecz3k (approved):

godot_project/editor/controls/workspace_view/tab_bar_active_workspaces.gd
MarianoGnu: yes

PR #42, opened 2024-11-04
Fix missing outline when closing the screen capture popup

HungryProton:
The selection state was properly saved (and restored), but the rendering state was not. Before the snapshot system, changing the selection used to cause the renderer to refresh its state, but now this has to be done explicitly.

### DIFF
--- a/godot_project/project_workspace/workspace_context/structure_context/selection_db/selection_db.gd
+++ b/godot_project/project_workspace/workspace_context/structure_context/selection_db/selection_db.gd
@@ -459,55 +459,5 @@ func apply_selection_snapshot(in_snapshot: Dictionary) -> void:
 	var atom_snapshot: Array = in_snapshot.atom_snapshot
 	var spring_snapshot: Array = in_snapshot.spring_snapshot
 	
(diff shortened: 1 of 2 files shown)


godot_project/project_workspace/workspace_context/structure_context/selection_db/selection_db.gd
```diff
@@ -459,55 +459,5 @@ func apply_selection_snapshot(in_snapshot: Dictionary) -> void:
...
-		# atom_id<int> : is_atom_currently_selected<bool>
-	}
-	
-	var bond_to_selection_status: Dictionary = {
-		# bond_id<int> : is_bond_currently_selected<bool>
-	}
-	var spring_to_selection_status: Dictionary = {
-		# spring_id<int> : is_spring_selected<bool>
-	}
-
```
HungryProton, line 472: This is dead code that should have been removed when merging the snapshot system.

It was there when we had to manually call `representation.highlight_atoms()` after changing a selection, but that's not the case anymore.

kubecz3k (approved):

PR #43, opened 2024-11-04
fix adding a structure after simulation

kubecz3k:
CRASH
-----------
fix Rendering snapshot edge case issue 🔨

There was an issue related to snapshot implementation of Rendering that
revealed itself in case when snapshot has been applied multiple times
during the same frame.
In case when in the result of snapshot 'apply' operation the structure
has been removed, and then later during the same frame other snapshot
'apply' operation is requested this time re-instantiating the same
structure, such structure was not re-instantiated correctly.

The direct cause for this was incomplete is_instance_valid check(),
such instance is valid at that point but already queued for deletion.

The edge case described in here is happening every time when directly
after running mechanical simulation user decides to add new structure
from library (WorkspaceContext.snapshot_moment() performs couple
apply_snapshot in a row in such case)

### DIFF
--- a/godot_project/project_workspace/workspace_context/workspace_context.gd
+++ b/godot_project/project_workspace/workspace_context/workspace_context.gd
@@ -576,11 +576,9 @@ func get_nano_structure_context(in_nano_structure: NanoStructure) -> StructureCo
 	var guid: int = in_nano_structure.int_guid
 	assert(workspace.has_structure(in_nano_structure))
 	if !_structure_contexts.has(guid):
-		var structure_context: StructureContext = null
-		if structure_context == null:
-			structure_context = StructureContextScn.instantiate()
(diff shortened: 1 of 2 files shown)


godot_project/project_workspace/workspace_context/workspace_context.gd
```diff
@@ -576,11 +576,9 @@ func get_nano_structure_context(in_nano_structure: NanoStructure) -> StructureCo
 	var guid: int = in_nano_structure.int_guid
 	assert(workspace.has_structure(in_nano_structure))
 	if !_structure_contexts.has(guid):
-		var structure_context: StructureContext = null
-		if structure_context == null:
```
kubecz3k, line 580: this check was redundant

HungryProton (approved):

PR #51, opened 2024-11-05
Support to edit position of virtual objects in Selection Info Panel of the Dynamic Context docker

MarianoGnu:
----------
 Tasks:
- Part 1 of BUG: Position and Rotation of Reference shapes are not editable in Selection Information panel
- Part 1 of BUG: Virtual Motors do not show any information in SelectionInformation panel
- BUG: Selected VirtualAnchor and Springs show not information at all in Selection Information panel


kubecz3k (approved):
Don't see anything suspicious + tested locally and can confirm I'm able to change position for anchors/motors in selection panel

PR #53, opened 2024-11-06
Add support to editing rotation of virtual objects in Selection Info Panel of the Dynamic Context docker

MarianoGnu:
- NanoVirtualMotor has special treatment because their model is facing to the right instead of to the front
- Also modified the rules when positioning the editor controls to expand using both columsn when first column has no text (This is exactly the case of Position and Rotation)

-----------
 Tasks:
- Part 2 of BUG: Position and Rotation of Reference shapes are not editable in Selection Information panel
- Part 2 of BUG: Virtual Motors do not show any information in SelectionInformation panel

### DIFF
--- a/godot_project/project_workspace/workspace_context/structure_context/selection_info.gd
+++ b/godot_project/project_workspace/workspace_context/structure_context/selection_info.gd
@@ -1,6 +1,7 @@
 class_name SelectionInfo extends Object
 
 const InspectorControlVector3Scene = preload("res://editor/controls/dockers/workspace_docker/c_dynamic_context_docker/dynamic_context_controls/inspector_controls/inspector_control_vector3/inspector_control_vector3.tscn")
+const InspectorControlRotationScene = preload("res://editor/controls/dockers/workspace_docker/c_dynamic_context_docker/dynamic_context_controls/inspector_controls/inspector_control_rotation/inspector_control_rotation.tscn")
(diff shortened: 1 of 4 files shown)


godot_project/project_workspace/workspace_context/structure_context/selection_info.gd
```diff
@@ -337,3 +358,74 @@ class SetVirtualObjectPositionHelper:
...
+		changed_signal = out_structure_context.nano_structure.transform_changed
+		_structure_context = out_structure_context
+	
+	
+	func get_rotation_euler() -> Vector3:
+		var basis: Basis = _structure_context.nano_structure.get_transform().basis
+		var euler: Vector3 = basis.get_euler()
+		if _type == Type.MOTOR:
+			# Virtual Motors faces RIGHT instead of FORWARD
+			# because of this we need a special conversion:
```
HungryProton, line 390: What implications would it have to make the virtual motors face forward like everything else instead?

HungryProton (approved):

godot_project/project_workspace/workspace_context/structure_context/selection_info.gd
kubecz3k: Might be wrong but I think it's matter of rotating model in blender and exporting it again, seems like a bit more elegant solution

MarianoGnu: if we change the model we have to update the OpenMM simulation to know the new axis, no big deal. But my concern is loading an already created file will put the motors pointing to the wrong direction

PR #57, opened 2024-11-07
Fix Bonds disconnected from atoms during rotation

kubecz3k:
Bonds disconnected from atoms during rotation
-----------

    fix bond independent movement after atom creation 🔨
    
    When user had non selected bond connecting two selected atoms and they
    created new atom, the bond moved together with newly created atom
    despite the fact the related atoms were standing still (not the bond
    nor related atoms were selected at that point)
    
    This issue was caused by the fact in such case shader never gets
    information that the bond is no longer connected to any selected atom
    This commit fixes the issue by ensuring SelectionDB.set_atom_selection
    always refreshes any involved bond

    fix bond independent movement after deselection 🔨
    
    When user had non selected bond connecting two selected atoms and they
    deselected those atoms one by one, then after doing drag drop selection
    the bond moved twice as fast as it should.
    
    It was caused by error in StickRepresentation where 'partial_selection'
    variable value was not representing real situation anymore (missing
    check for case where bond is no longer influenced by any atom)
    As a result of bond being inside this variable it has been moved twice,
    once by cpu and then by the shader

To anyone who will review this, it's probably the best to look at one commit at a time for simplicity

### DIFF
--- a/godot_project/editor/rendering/atomic_structure_renderer/representation/stick_representation/stick_representation.gd
+++ b/godot_project/editor/rendering/atomic_structure_renderer/representation/stick_representation/stick_representation.gd
@@ -156,7 +156,7 @@ func highlight_atoms(_in_atoms_ids: PackedInt32Array, \
 func _refresh_bond_partial_influence_status(new_partially_influenced_bonds: PackedInt32Array) -> void:
 	var structure_context: StructureContext = _workspace_context.get_structure_context(_related_structure_id)
 	var related_structure: AtomicStructure = structure_context.nano_structure as AtomicStructure
-	for bond_id in new_partially_influenced_bonds:
+	for bond_id: int in new_partially_influenced_bonds:
 		if not bond_id in _bond_id_to_particle_id:
 			continue
 		var bond: Vector3i = related_structure.get_bond(bond_id)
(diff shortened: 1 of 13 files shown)


godot_project/editor/rendering/atomic_structure_renderer/representation/stick_representation/stick_representation.gd
```diff
@@ -196,9 +198,6 @@ func lowlight_atoms(_in_atoms_ids: PackedInt32Array, in_bonds_released_from_part
 			new_partially_influenced_bonds: PackedInt32Array = PackedInt32Array()) -> void:
 	_refresh_bond_partial_influence_status(in_bonds_released_from_partial_influence)
 	_refresh_bond_partial_influence_status(new_partially_influenced_bonds)
-	if _current_bond_partial_selection.size() == in_bonds_released_from_partial_influence.size():
-		_current_bond_partial_selection.clear()
```
kubecz3k, line 200: this actually lead to third issue which I haven't documented, this assumed that since there is the same nmb of bonds that just have been released from partial influence as there have been in the past those two collections are the same. 
I had edge case where it's not true (for case where both collections have size = 1)

This check is totally unneeded after changes from 184-188 lines

godot_project/editor/rendering/atomic_structure_renderer/representation/stick_representation/stick_representation.gd
```diff
@@ -584,10 +583,13 @@ static func _calc_up_vector_for_higher_bond(in_first_atom_id: int, in_second_ato
...
-	_refresh_bond_partial_influence_status(_current_bond_partial_selection.keys())
-	_current_bond_partial_selection.clear()
-	_refresh_bond_partial_influence_status(in_partially_selected_bonds)
+func refresh_bond_influence(in_partially_selected_bonds: PackedInt32Array) -> void:
+	var bonds_to_refresh: Dictionary = {}
+	for bond_id: int in in_partially_selected_bonds:
+		bonds_to_refresh[bond_id] = true
+	for bond_id: int in _current_bond_partial_selection:
+		bonds_to_refresh[bond_id] = true
+	_refresh_bond_partial_influence_status(bonds_to_refresh.keys())
```
kubecz3k, line 592: Those changes do not fix anything, this change have only one goal in mind, to avoid potential call of _refresh_bond_partial_influence_status logic twice for the same bond

PR #59, opened 2024-11-08
fix non-existing function call 🔨

kubecz3k:
"Bonds disconnected from atoms during rotation"

-----------
The function has been renamed recently, but it was not updated everywhere


kubecz3k:
Merging proactively since changes are banal and this prevents crashes, any eventual feedback will be addressed in subsequent pr

PR #66, opened 2024-11-10
fix double bond preview rotation 🔨

kubecz3k:
"Drag-drop double bond looks like single bond"
-----------
fix double bond preview rotation 🔨
    
    There has been a problem with double bond preview, from user
    perspective it looked exactly like single bond preview.
    It was caused by improper rotation that always positioned that mesh to
    be seen 'from a side'.


MarianoGnu (approved):
ship it

PR #68, opened 2024-11-11
Fix msep1 file load crash  🔨

kubecz3k:
"Crash when loading msep1 file from directory without write permissions"
-----------
This commit ensures user is not experiencing crash when trying to load msep1 file without write permission
This change should also serve any problem with file corruption


MarianoGnu (approved):
LGTM

PR #69, opened 2024-11-11
Prevent local change in font file after reimporting assets

kubecz3k:
"Problem with Gidole-Regular.ttf import"
-----------
After deleting .godot directory and reimporting assets the Gidole-Regular font is being changed by adding blank empty line at the end
It seems it's no longer happening after accepting this change


kubecz3k:
There is still similar problem with MotorLink_Mesh.res

MarianoGnu (approved):

PR #70, opened 2024-11-11
Make the preview of the small molecule appear as soon as finishes loading, even if mouse doesn't move

MarianoGnu:
Task: BUG: When switching from one small molecule to another, the preview will remain invisible until mouse moves

### DIFF
--- a/godot_project/editor/controls/dockers/workspace_docker/a_create_docker/create_fragment_panel.gd
+++ b/godot_project/editor/controls/dockers/workspace_docker/a_create_docker/create_fragment_panel.gd
@@ -88,6 +88,18 @@ func _on_fragment_selected(fragment_path: String) -> void:
 	var structure: NanoStructure = await WorkspaceUtils.get_nano_structure_from_file(workspace_context, absolute_path, false, false, false)
 	structure.set_structure_name(fragment_path.get_file().get_basename())
 	workspace_context.create_object_parameters.set_new_structure(structure)
+	# HACK: for responsivemes, we will "fake" a mouse movement to ensure preview of the molecule will
+	# reappear in the desired position only if the mouse is hovering the viewport
+	if BusyIndicator.visible:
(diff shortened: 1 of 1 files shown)


godot_project/editor/controls/dockers/workspace_docker/a_create_docker/create_fragment_panel.gd
```diff
@@ -88,6 +88,18 @@ func _on_fragment_selected(fragment_path: String) -> void:
...
+	# reappear in the desired position only if the mouse is hovering the viewport
+	if BusyIndicator.visible:
+		# wait for busy indicator to fully dissapear, otherwise will consume the event
+		await BusyIndicator.visibility_changed
+		await get_tree().process_frame
+	var main_viewport: Viewport = get_tree().root
+	var mouse_move := InputEventMouseMotion.new()
+	mouse_move.position = main_viewport.get_mouse_position()
+	mouse_move.relative = Vector2.ZERO
+	main_viewport.push_input(mouse_move)
```
kubecz3k, line 101: nit: since it's rather hacky, and it's using await I would move it to separate private method

kubecz3k (approved):
I'm a bit on a fence if introducing this hack is worth the effect, never heard any complain about this behavior, the part that makes me the most nervous is introducing new await, which is something I'm always trying to avoid since those sometimes backfires in unexpected way. 
That being said this indeed fixes the issue and is easy to understand so I guess good to merge

godot_project/editor/controls/dockers/workspace_docker/a_create_docker/create_fragment_panel.gd
MarianoGnu: honestly, i tough about it and i dont think it would help, it would add confusion by making people believe there may be a need of reusing the function, which at the moment is not true. Also i would not know how to name it

PR #71, opened 2024-11-11
"End Simulation" feature

HungryProton:
Adds a "End Simulation" button to the Molecular Simulations panel.

Pressing it stops processing the simulation on OpenMM, but keeps the simulation data to that point in MSEP.


godot_project/editor/controls/dockers/workspace_docker/simulations_docker/simulations_docker_controls/simulation_tools_panel.gd
```diff
@@ -272,6 +276,7 @@ func _update_controls() -> void:
 			_spin_box_timeline.editable = true
 			_slider_timeline.editable = true
 			_button_revert.disabled = false
+			_button_end.disabled = true
```
MarianoGnu, line 280: we are missing one case to set `disabled = true` that is when the simulation finishes procesing
```
func _on_simulation_frame_received(in_time: float, _in_state: Variant) -> void:
	_simulation_length_nanoseconds = max(in_time, _simulation_length_nanoseconds)
+	if _simulation_length_nanosecinds == _total_simulation_len_nanoseconds:
+		_button_end.disabled = true
```

MarianoGnu (approved):
Approved with one comment

godot_project/editor/controls/dockers/workspace_docker/simulations_docker/simulations_docker_controls/simulation_tools_panel.gd
HungryProton: Good catch!

PR #72, opened 2024-11-11
When an error happens while assigning potential charge to molecules, convert atom_id in the error to one MSEP can understand

MarianoGnu:
-----------
 Task: BUG: When trying to relax this file, error referrs to an invalid valence in Nitrogen, but wrong atom (an Oxygen) is selected instead
[all-small-molecules.zip](https://github.com/user-attachments/files/17706032/all-small-molecules.zip)

### DIFF
--- a/godot_project/python/scripts/openmm_server.py
+++ b/godot_project/python/scripts/openmm_server.py
@@ -19,6 +19,10 @@ import logging
 
 DETAILED_LOGS = False
 
+try:
+	os.mkdir(os.path.join(os.path.dirname(os.path.realpath(__file__)), "..", "..", "logs" ))
+except Exception as e:
+	pass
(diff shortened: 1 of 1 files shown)


godot_project/python/scripts/openmm_server.py
```diff
@@ -19,6 +19,10 @@
 
 DETAILED_LOGS = False
 
+try:
+	os.mkdir(os.path.join(os.path.dirname(os.path.realpath(__file__)), "..", "..", "logs" ))
```
kubecz3k, line 23: this one fixes some other issue?

kubecz3k (approved):

godot_project/python/scripts/openmm_server.py
MarianoGnu: i got an exception when running from the debugger, so i added this guard 🤷

PR #76, opened 2024-11-12
Clean up 'color snapshot' semantics 🔧

kubecz3k:
'Color snapshot' nomenclature seems misleading after introduction of snapshot api
This commit removes unused function and renames the other one in such way it should not get confused with what snapshots are


kubecz3k:
Note: the "#TODO: This is probably not needed (THIS COMMENT SHOULD NOT BE PART OF PR)" comments were created during snapshot implementation

HungryProton (approved):

PR #84, opened 2024-11-12
Show 'Help > About MSEP.one' menu item in MacOS

MarianoGnu:
It was hidden because it was already accesible from the Application Menu, but removing it causes a regression where 'Tutorials and Documentation' doesn't work on this platform

Task: BUG - Tutorial PDF doesn't launch **ON MACOS** when selected from the menu bar pulldown menu
-----------


brunosxs (approved):
LGTM

PR #85, opened 2024-11-13
fix selection shrink 🔨

kubecz3k:
"Shrink selection deselects all"
-----------
This commit fixes selection shrink, before the shrink simply deselected all


kubecz3k:
I probably won't be around, so if this gets approved I will appreciate a merge before the build will be made

MarianoGnu (approved):
Works properly. Sheep it :sheep:

PR #88, opened 2024-11-14
Select source atom when performing drag-drop 🔧

kubecz3k:
"Preview Atom Can be Wrong Size in Perspective mode"
-----------
    select source atom when performing drag-drop 🔧
    
    This commit ensures the source atom of drag and drop is selected.
    It's done in order to ensure preview is being visualized at creation
    distance which is expected by the user
    (in case when selected atom was at significantly different position
    from drag-drop source atom the preview was visualized nearby that
    selection)


brunosxs (approved):
LGTM

Issue #95, opened 2024-12-16
'ascii' codec can't decode byte 0xe2 in position 1467: ordinal not in range(128) When running Relaxation


poppingtonic:
When running 'Relaxation' on a carbon 5-ring after correcting Hydrogens.

![Screenshot from 2024-12-16 23-27-47](https://github.com/user-attachments/assets/f176a40d-09b2-4dfd-a497-df75c292f452)

```  
Traceback:

    File openmm_server.py:1235, in <module>
    1235|   minimized_positions = minimize_energy(topology, state, temperature_in_kelvins, max_iterations=500)
 File openmm_server.py:807, in minimize_energy
    807|   forcefield = create_forcefield_for_topology(topology_payload)

    File openmm_server.py:800, in create_forcefield_for_topology
    800|   forcefield.parse_sources([forcefield_extension_path])

    File <env>/lib/python3.10/site-packages/openff/toolkit/typing/engines/smirnoff/forcefield.py:795, in parse_sources
    795|   smirnoff_data = self.parse_smirnoff_from_source(source)
 File <env>/lib/python3.10/site-packages/openff/toolkit/typing/engines/smirnoff/forcefield.py:1016, in parse_smirnoff_from_source
    1016|   smirnoff_data = parameter_io_handler.parse_file(source)

    File <env>/lib/python3.10/site-packages/openff/toolkit/typing/engines/smirnoff/io.py:129, in parse_file
    129|   raw_data = source_obj.read()

    File <env>/lib/python3.10/encodings/ascii.py:26, in decode
    26|   return codecs.ascii_decode(input, self.errors)[0]
```

PR #97, opened 2024-12-24
Fix: Specify utf-8 encoding to ensure proper handling of extended characters

poppingtonic:
'ascii' codec can't decode byte 0xe2 in position 1467: ordinal not in range(128) #95 
-----------
 
The error occurred because Python was trying to read the file using ASCII encoding by default, but encountered a non-ASCII character (byte 0xe2) which couldn't be decoded.

This is a common issue that can occur specifically on Linux because different operating systems have different default file encodings:
- Windows typically defaults to cp1252 (a superset of ASCII)
- macOS typically defaults to UTF-8
- Linux's default can vary, but often defaults to ASCII

So we explicitly set `utf-8` as the default encoding.


hpvb:
Thank you!

PR #99, opened 2025-01-27
Fix the auto pose recommendations and bond order

HungryProton:
This PR forces the auto pose feature to always use the selected atomic element, instead of switching to oxygen or hydrogens based on the parent atom free valence.

Clicking a candidate will now pick a smaller bond order if necessary.

### DIFF
--- a/godot_project/autoloads/data/ElementData.gd
+++ b/godot_project/autoloads/data/ElementData.gd
@@ -75,4 +75,3 @@ func _get_contact_radius() -> float:
 		return contact_radius
 	var fallback_factor: float = ProjectSettings.get_setting(VAN_DER_WAALS_FALLBACK_FACTOR_SETTING, 1.5)
 	return render_radius * fallback_factor
-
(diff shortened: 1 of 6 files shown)


godot_project/autoloads/data/ElementData.gd
```diff
@@ -76,3 +76,9 @@ func _get_contact_radius() -> float:
 	var fallback_factor: float = ProjectSettings.get_setting(VAN_DER_WAALS_FALLBACK_FACTOR_SETTING, 1.5)
 	return render_radius * fallback_factor
 
+
+func get_maximum_bonds_count() -> int:
+	var valence_left: int = valence
+	if number > 5:
+		valence_left = 8 - valence_left
+	return valence_left
```
MarianoGnu, line 84: this method is never used and is also wrong, as an example valence of helium is 2, but bonds count of helium is meant to be 0

MarianoGnu (approved):
Address comment before merging

godot_project/autoloads/data/ElementData.gd
HungryProton: Right I forgot to remove that
But if this is wrong, then this one is wrong too I guess? https://github.com/MSEP-one/msep.one/blob/main/godot_project/project_workspace/structs/atomic_structure.gd#L343

MarianoGnu: we could try to correct the hydrogens on helium and see what happens

PR #102, opened 2025-02-12
Fix single atoms not visible in stick view

HungryProton:
The single atom representation shader is using a different billboard code that's not working, making unconnected atoms invisible in stick view.

This PR moves the billboard code (used by sticks, labels and sphere views) under a common shader include.
It's no longer necessary to pass the camera basis vector as uniform because that data is included in INV_VIEW_MATRIX already, so code related to that was removed.


brunosxs (approved):
lgtm

PR #103, opened 2025-02-12
Fix atom suggestions not being restored after moving the camera with the keyboard

HungryProton:
The atom suggestions wouldn't re appear until the next mouse event if the camera was moved with the keyboard.

This is because a wrong condition in handle_inputs_resume.


brunosxs (approved):
lgtm

PR #104, opened 2025-02-13
Auto posing freezes MSEP when loading a large file

HungryProton:
There are two main performance issues with the atom auto posing feature:

+ Drawing 100+ circles, connections and labels is expensive
+ When two candidates are close to each other, they are merged into a single one, connecting all the parent atoms.
  - The current implementation of this check is O(n²)
  - This is what causing the several seconds freeze when loading a file or making a large selection at once.

This PR limits the maximum amount of candidates on screen (computed and drawn).
It also takes the spatial hash grid from the "atoms overlap detection" and moves it to a generic class. This way we can reuse it to detect the candidates to merge and speed up that step.

### DIFF
--- a/godot_project/editor/controls/dockers/workspace_docker/simulations_docker/simulations_docker_controls/atomic_structure_model_validator.gd
+++ b/godot_project/editor/controls/dockers/workspace_docker/simulations_docker/simulations_docker_controls/atomic_structure_model_validator.gd
@@ -47,7 +47,7 @@ func _validate_bonds_in_thread(
 			continue
 		var atomic_structure: AtomicStructure = structure_context.nano_structure as AtomicStructure
 		var ignored_springs: Array[Metadata] = []
-		var spatial_hash_grid: SpatialHashGrid = SpatialHashGrid.new(MAX_COVALENT_RADIUS)
+		var spatial_hash_grid: SpatialHashGridOverlaps = SpatialHashGridOverlaps.new(MAX_COVALENT_RADIUS)
 		var atoms: PackedInt32Array
 		if in_selection_only:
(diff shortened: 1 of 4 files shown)


godot_project/editor/controls/dockers/workspace_docker/simulations_docker/simulations_docker_controls/atomic_structure_model_validator.gd
```diff
@@ -549,82 +548,31 @@ class InvalidSp123Data extends Metadata:
 		return false
 
 
-## The spatial hash grid is used to partition the atoms in a 3D grid and speed
```
HungryProton, line 552: This part was moved to spatial_hash_grid.gd

godot_project/editor/input_handlers/molecular_structures/add_posing_atom_input_handler.gd
```diff
@@ -132,50 +133,45 @@ func set_preview_position(_in_position: Vector3) -> void:
...
-				var candidates_positions: PackedVector3Array = _generate_candidates_for_atom(context, atom_id)
-				var free_valences: int = context.nano_structure.atom_get_remaining_valence(atom_id)
-				for pos: Vector3 in candidates_positions:
-					var candidate := AtomCandidate.new()
-					candidate.structrure_id = context.nano_structure.int_guid
-					candidate.atom_ids = [atom_id]
-					candidate.atom_position = pos
-					candidate.total_free_valence = free_valences
-					structure_candidates.push_back(candidate)
-			# Merge close atoms
```
HungryProton, line 150: The following is the O(n²) part that's replaced with a hash grid

PR #105, opened 2025-02-19
Add video tutorials

HungryProton:
This PR adds: 
+ A dialog to browse tutorials
+ A floating video player
+ 2 video tutorials for MSEP

The specification requested to display a progress bar in the video player to control, but if the video stream doesn't support it (like the built-in Godot theora stream we're currently using), the bar is replaced with a Replay button.


HungryProton:
Waiting for feedback from Jonathan given the codec limitations.

PR #109, opened 2025-02-28
CRASH FIX: Ensure virtual objects and groups are marked 'dirty' when visibility is changed

MarianoGnu:
Fixes a crash that occurs when box selecting a hidden virtual object (anchor, or motor)

Task: CRASH - Invisible objects cause crash errors


MarianoGnu:
targeting the wrong branch

PR #110, opened 2025-02-28
CRASH FIX: Ensure virtual objects and groups are marked 'dirty' when visibility is changed

MarianoGnu:
Fixes a crash that occurs when box selecting a hidden virtual object (anchor, or motor)

Task: CRASH - Invisible objects cause crash errors

Before:

https://github.com/user-attachments/assets/e923f8c8-17e5-4b28-802c-0e4c324c4cf2

After:

https://github.com/user-attachments/assets/f25ae573-ab83-4b0c-a08b-857645bc67c9


HungryProton (approved):
LGTM

PR #112, opened 2025-02-28
Fix invalid candidates appearing on screen, not connected to any atoms

HungryProton:
This happens if you have atom candidates on screen, and select too many atoms at once. The auto posing feature gets disabled but the previous candidates were not cleared, so they stayed up on screen.

### DIFF
--- a/godot_project/editor/input_handlers/molecular_structures/add_posing_atom_input_handler.gd
+++ b/godot_project/editor/input_handlers/molecular_structures/add_posing_atom_input_handler.gd
@@ -136,6 +136,7 @@ func _update_candidates_if_needed() -> void:
 	for context: StructureContext in selected_contexts:
 		total_atoms_selected += context.get_selected_atoms().size()
 	if total_atoms_selected > MAX_ATOMS_FOR_AUTO_POSING:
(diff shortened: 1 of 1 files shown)


godot_project/editor/input_handlers/molecular_structures/add_posing_atom_input_handler.gd
```diff
@@ -130,6 +130,7 @@ func _update_candidates_if_needed() -> void:
 	if not _candidates_dirty:
 		return
 	_candidates.clear()
+	_get_rendering().atom_autopose_preview_set_candidates(_candidates)
```
MarianoGnu, line 133: wouldn't be better to move this line of code to line 140 right before the early return to avoid calling this method 2 times in the case a valid collection of candidates is set?

MarianoGnu (changes requested):

godot_project/editor/input_handlers/molecular_structures/add_posing_atom_input_handler.gd
HungryProton: Oh right, absolutely

MarianoGnu:
🐑 sheep it

MarianoGnu (approved):

PR #115, opened 2025-03-03
Restore some of the old state when loading a saved project

MarianoGnu:
- Zoom level in orthogonal projection will be restored
- Currently edited group will be restored

Tasks:
BUG - Ortogonal Camera is not restored properly after saving and restoring a file
BUG - Save files not maintaining UI position in groups


HungryProton (approved):
LGTM

PR #117, opened 2025-03-04
Adds star gears project to the library

HungryProton:
Star gears is an MSEP project with motors and springs but the library only supports PDB files (atoms & bonds only). 

This PR adds the ability to import full MSEP workspaces to the current workspace.


MarianoGnu (commented):
Why is there both `star_gears.msep1` and `star_gears.msep1.tres` files? they are meant to be the same file, and tres is only a temporary file created during loading and save

godot_project/utils/workspace_utils.gd
```diff
@@ -712,6 +715,90 @@ static func _is_invalid_mol_file(in_path: String) -> bool:
...
+						placement: ImportFileDialog.Placement, snapshot_name: String) -> void:
+	var imported_workspace: Workspace = ResourceLoader.load(path, "", ResourceLoader.CACHE_MODE_IGNORE)
+	if not imported_workspace:
+		Editor_Utils.get_editor().prompt_error_msg("Cannot load file: " + path)
+		return
+	
+	# Clear selection in other contexts
+	var other_contexts: Array[StructureContext] = out_workspace_context.get_structure_contexts_with_selection()
+	for context: StructureContext in other_contexts:
+		context.clear_selection()
```
MarianoGnu, line 735: nit: you can replace all this block for
`out_workspace_context.clear_all_selection()`

MarianoGnu (approved):
Because the load process generates a `.tres` file in the same location as the `.msep1` file we should test how this works after export, because the template file will reside inside of the MSEP.pck file.

The pdb/mol importer creates a copy of the file in a temp folder before actually importing it

HungryProton:
> Because the load process generates a `.tres` file in the same location as the `.msep1` file we should test how this works after export, because the template file will reside inside of the MSEP.pck file.
> 
> The pdb/mol importer creates a copy of the file in a temp folder before actually importing it

You're right, that doesn't work in exported builds. I'll see if I can move the tres file in another location.

MarianoGnu:
Lets address in the next PR, Jonathan is asking to merge stuff

PR #118, opened 2025-03-04
Fixes candidate recommendations collisions

HungryProton:
Increased the minimum distance from a candidate to existing atoms.

The `MAX_DISTANCE_TO_ATOMS` was misleading and was renamed to `MIN_DISTANCE_TO_ATOMS`


MarianoGnu (commented):
I have no strong opinions agains or in favor of this PR, buthis does not solve what is reported in the Task.

This is the scenario that is reported:
image

You cannot click in the candidate that is on top of an atom because it will select the atom instead of creating the candidate

PR #126, opened 2025-03-06
Fix star gears missing connections

HungryProton:
Fixes: "Add Star Gears Project to Sample Library - The file doesn’t seem to import correctly. Neither of the motors are attached to groups"

Springs, anchors and motors were pointing at the old structures id (from before the import), causing various issues. 

This PR adds a new `init_remap_structure_ids()` method in `NanoStructure`, called automatically when a workspace is imported from another workspace. If the structure is holding a reference to another structure, it has to update that link here.

### DIFF
--- a/godot_project/project_workspace/workspace.gd
+++ b/godot_project/project_workspace/workspace.gd
@@ -66,8 +66,6 @@ static var instance_counter: int = 0
 	set = _set_representation_settings
 
 
-
-
 ## User defined simulation settings
 ## like simulation duration, time step, relax before simulating, etc... stored inside the project
 ## These are updated in the moment the user starts a new simulation, not before
@@ -145,6 +143,7 @@ var suggested_path: String = ""
 
 var _main_structure_int_guid: int = 0
 var _id: int
+var _reserved_int_guids: Dictionary = {} # {guid<int>: true}
 
 
 func _init() -> void:
@@ -162,12 +161,13 @@ func _init() -> void:
 func post_load() -> void:
 	if _main_structure_int_guid != INVALID_STRUCTURE_ID:
 		return
+	# Rebuild the reserved id list
 	# Main structure is the only one without a parent
 	for structure_id: int in _structures:
 		var structure: NanoStructure = _structures[structure_id]
+		_reserved_int_guids[structure.int_guid] = true
 		if structure.int_parent_guid == INVALID_STRUCTURE_ID:
 			_main_structure_int_guid = structure_id
-			return 
 
 
 func get_nmb_of_structures() -> int:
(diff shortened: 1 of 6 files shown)


godot_project/project_workspace/workspace.gd
```diff
@@ -345,16 +347,17 @@ func is_a_ancestor_of_b(in_ancestor_candidate: NanoStructure, in_descendant_cand
...
 	var int_id: int = 0
 	while !is_valid:
 		# int_id must be in 32 bit range since we are using PackedInt32Array across the project
 		int_id = _rng.randi_range(0, MAX_SIGNED_32_BIT_INT)
 		
-		if int_id <= 0 || has_structure_with_int_guid(int_id):
+		if int_id <= 0 || _reserved_int_guids.has(int_id):
 			continue
 		is_valid = true
+	_reserved_int_guids[int_id] = true
```
MarianoGnu, line 360: out of curiosity, why was this varuable introduced? since this is not an @export variable, this variable will not survive being saved and reloaded, and if i am not mistaken `_internal_add_structure` is not called again every time for every structure when the file is loaded, so `_reserved_int_guids` is not properly populated when loading a saved file, risking the unlikely case in which RNG will generate twice the same 32 bits unsigned number and 2 structures will have the same ID
⚠️

MarianoGnu (changes requested):

godot_project/project_workspace/workspace.gd
HungryProton: True, this one should be an export, and probably part of the snapshot state too.

As for the reason the variable was added:

The imported structures can't reuse their old guid (from the imported file) or they will conflict when we import the same workspace twice. So all the structures needs a new GUID. Currently it's like this:

+ Load all the structures from the file
+ Sort from parent to leaves
+ For each structure to add:
  1. Reset the structure GUID
  2. Add the structure to the workspace (new GUID is generated here)
  3. Build the renderer and other stuff

Issue is: Motors, springs and anchors holds references to their linked structures, and some of these links are bidirectional.

And call `add_structure()` triggers other stuff that requires at least some of these links to be up to date, but that's also when they get they're new GUID.

So I moved the GUID creation to a public method, generates the new GUID before adding the structures to the workspace, update the links, and only then the structures are added to the workspace.

But maybe I missed something and there's a way to keep the GUID generation internal. Maybe by modifying add_structure() so all the asserts and renderer rebuilds are deferred, but that sounds like a whole different class of issues.

MarianoGnu: Yes, you can keep creation internal:
```
	var old_to_new_id_map = {}
		for structure: NanoStructure in structures_to_add[old_parent_id]:
			var old_id: int = structure.int_guid
			structure.int_guid = Workspace.INVALID_STRUCTURE_ID
			out_workspace_context.workspace.add_structure(structure)
			old_to_new_id_map[old_id] = structure.int_guid
```

HungryProton: That loop is how it's already done, and that's the issue. I need to know the new guid of all the structures, before the structures are added to the workspace

HungryProton: I've included the _reserved_int_guid in the snapshot, but I kept it as an internal variable instead of an imported one. The list is rebuilt on load and doesn't need to be saved on disk.

MarianoGnu: i dont think knowing the guids before adding them is important, just add all the structures, and latter loop them again to run the `init_remap_structure_ids` method

MarianoGnu (approved):

PR #129, opened 2025-03-10
Dont prevent from closing the application when there are open dialogs

MarianoGnu:
Task: UI Bug - When in Full Screen video playback mode, the minimize button shouldn't read "Full Screen"


godot_project/editor/input_handlers/molecular_structures/dialog_input_handler.gd
```diff
@@ -0,0 +1,27 @@
+extends InputHandlerBase
```
MarianoGnu, line 1: This inut handler has the same job the `exclusive` property used to do, preventing changes on the project while dialogs are open. However it will not prevent users from using the side panel

HungryProton (approved):

PR #135, opened 2025-03-12
Fix `import workspace` logic & code cleanup

HungryProton:
Import workspace now works with any group hierarchy, including when an child anchor is connected to atoms from an ancestor structure.

+ Fixes anchor not respecting the selected placement options
+ `_create_in_guid` from Workspace is now internal again
+ Code should be easier to follow


MarianoGnu (commented):
LGTM :D

MarianoGnu (approved):

PR #136, opened 2025-03-12
Change behaviour of 'Select atoms by type' panel. Will select only atoms per group

MarianoGnu:
- This is for consistency, previously it allowed to have subgroups partially selected, something that is impossible to do any other way
- When clicking on a group listed in the new UI element, it will make that group active, and select the atoms only of that group
- If the group clicked doesn't have atoms, then inout will be ignored and no change on active group or selection will be made

**Known Issue**
- Right now hidden atoms will not be made possible. Waiting for input from @jmackley100 to decide what should happen to them and/or how to warn about their existence to user

Task: BUG - Select Atom by Type allows partial selection of subgroups


HungryProton (approved):
Looks good to me, the only things I would change (maybe in another PR) is:
+ Ignore the hidden atoms and don't show them in the count
+ Highlight the clicked tree item (and remove the highlight effect when the selection changes from outside) 

Or maybe we could add a "Include hidden atoms" checkbox at the top, and that would automatically show the relevant hidden atoms. I don't know if Jonathan gave your more info about what should be done?

PR #146, opened 2025-03-28
Fix bugs related to cutting groups

HungryProton:
Fixes:
+ CRASH - Cut Group
+ CRASH BUG - Cut and Paste doesn't remember Group Name

The structure name is now stored in the clipboard and reused when available
The crashes were caused by the sphere representation trying to update the same frame the AtomicRenderer was queued free.


godot_project/export_presets.cfg
hpvb, line 1: Why are there changes to the export presets in this PR?

HungryProton: Oh right, these changes aren't supposed to be here

PR #158, opened 2025-04-09
Show a warning message when using the auto posing feature with more than 50 atoms

HungryProton:
Fixes: "BUG - Atom Placement Hints disappear when using select all"

### DIFF
--- a/godot_project/editor/input_handlers/molecular_structures/add_posing_atom_input_handler.gd
+++ b/godot_project/editor/input_handlers/molecular_structures/add_posing_atom_input_handler.gd
@@ -86,6 +86,7 @@ func forward_input(in_input_event: InputEvent, _in_camera: Camera3D, out_context
 		_ensure_create_mode()
 	
 	update_preview_position()
+	_update_candidates_if_needed()
 	rendering.atom_autopose_preview_show()
 	
 	if in_input_event is InputEventMouse:
@@ -147,7 +148,7 @@ func set_preview_position(_in_position: Vector3) -> void:
 
 
 func _update_candidates_if_needed() -> void:
-	if not _candidates_dirty:
+	if not _candidates_dirty or not _should_show():
 		return
 	_candidates.clear()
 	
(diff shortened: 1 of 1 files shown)


godot_project/editor/input_handlers/molecular_structures/add_posing_atom_input_handler.gd
```diff
@@ -86,6 +86,7 @@ func forward_input(in_input_event: InputEvent, _in_camera: Camera3D, out_context
 		_ensure_create_mode()
 	
 	update_preview_position()
+	_update_candidates_if_needed()
```
MarianoGnu, line 89: What's the intention of updating candidates when candidates have not changed? you should not need to update the message unless selection have changed again, this looks like unnecesary workload to me 🤔

godot_project/editor/input_handlers/molecular_structures/add_posing_atom_input_handler.gd
```diff
@@ -355,6 +358,8 @@ func _should_show() -> bool:
 		return false
 	if parameters.get_create_mode_type() != CreateObjectParameters.CreateModeType.CREATE_ATOMS_AND_BONDS:
 		return false
+	if _is_shortcut_pressed():
+		return true
```
MarianoGnu, line 362: this is bypassing `_workspace_context.workspace.representation_settings.get_display_auto_posing()` i dont think that was  required or desired

MarianoGnu (changes requested):
Making some questions, i would like at least some answers before approval

godot_project/editor/input_handlers/molecular_structures/add_posing_atom_input_handler.gd
HungryProton: Currently, if you hold the alt key down, candidates will show up (if we're in create mode), regardless of was the display settings say. 
So, either the current behavior is wrong, either this line was missing to match the current behavior.

MarianoGnu: Well, that's probably a sepparate bug that needs to be resolved then

PR #163, opened 2025-04-17
Hide atoms suggestions during relaxation

HungryProton:
Fixes: BUG - Atom Position Recommendations should disappear during relaxation


godot_project/editor/input_handlers/molecular_structures/add_posing_atom_input_handler.gd
```diff
@@ -436,6 +438,15 @@ func _on_structure_contents_changed(structure_context: StructureContext) -> void
 		_atom_grid = null
 
 
+func _on_workspace_context_atom_relaxation_started() -> void:
+	_hide_preview()
+
+
+func _on_workspace_context_atoms_relaxation_finished(_error: String) -> void:
+	if _should_show():
+		_show_preview()
```
MarianoGnu, line 447: maybe `_should_show()` also needs to account for relaxation

If i recall correctly, right now the BusyScreen will take an screenshot before showing, and mantain the view, so if candidates appears it will still not affect the rendering, so i may be iverthinking this

MarianoGnu (approved):

godot_project/editor/input_handlers/molecular_structures/add_posing_atom_input_handler.gd
MarianoGnu: you could set a local variable `_is_relaxing = true/false` on the signals, and read that instead of polling it from workspace_context

PR #166, opened 2025-04-18
Changes in preparation for upgrading to Godot 4.4.1

MarianoGnu:
- ColorPalette enum had to be renamed to ColorSchema because the name is taken [by a new resource](https://docs.godotengine.org/en/4.4/classes/class_colorpalette.html) in Godot 4.4: 
- every mention of the variables holding this enum was also relamed to color_schema
- Godot 4 enables a warning when a signal is not being emitted in the same script where is declared, so new methods for emmiting them was created (there's one exception, where the signal was emitted by a subclass of the script)
- 3D Preview now hides atom autopose preview (this is an actual fix that came up while testing)
- Some more changes in initialization logic that responds to what seems to be a change in the order of initialization in nodes

Task: Part 1 of **Update MSEP to latest GODOT version.**


MarianoGnu:
Please test and confirm this still works in Godot 4.2

godot_project/project_workspace/custom_shapes/pyramid_mesh.gd
```diff
@@ -52,10 +52,6 @@ func get_height() -> float:
 	return height
 
 
-func request_update() -> void:
-	_update()
-
-
```
HungryProton, line 58: Deleting this method means it doesn't work on 4.2 anymore because it's called in the setters

HungryProton: Apart from that, this PR still works perfectly with 4.2 :+1:

MarianoGnu: right! i should move this to the next PR

HungryProton (approved):
LGTM

PR #172, opened 2025-05-05
Prevent a freeze when simulating a structure with incomplete valence

HungryProton:
Fixes: CRASH - Rotating Nested Groups

This is more of a band aid fix than a proper patch, as I can't figure out the root cause of the issue.
The simulation `start_promise` is fulfilled while the thread is still running, causing the `wait_to_finish()` call to freeze the application, but doing so also prevents the thread from finishing somehow.

I suspect there's some mutexes interlocking somewhere but this is bad enough that this fix might be necessary while we figure this out.


MarianoGnu (commented):
I will give a look to this issue before approving this, because this patch does not warrantee the ending of the thread

MarianoGnu (changes requested):
I am proposing #174 as an alternative to this one

PR #176, opened 2025-05-05
Enable etc2 texture compression required for macos export

HungryProton:
Export to MacOS is blocked if this setting is disabled.
Enabling it only for macos using the feature field in the export options doesn't work, so this is enabled project wide.

### DIFF
--- a/godot_project/project.godot
+++ b/godot_project/project.godot
@@ -346,6 +346,7 @@ limits/debugger/max_chars_per_second.editor=3000000
 
 [rendering]
 
+textures/vram_compression/import_etc2_astc=true
 textures/vram_compression/import_s3tc_bptc.macos=true
 
 [shader_globals]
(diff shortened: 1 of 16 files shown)


godot_project/project.godot
```diff
@@ -346,6 +346,7 @@ limits/debugger/max_chars_per_second.editor=3000000
 
 [rendering]
 
+textures/vram_compression/import_etc2_astc=true
 textures/vram_compression/import_s3tc_bptc.macos=true
```
HungryProton, line 350: I'm not sure this does anything anymore, and doing the same for etc2 doesn't work either.

MarianoGnu: this seems to always enable etc2, not only for macos. (Not that i bother)

Other than that the list of .import files affected by this change feels small. Before submitting please try to force reimporting all texture assets and see if there's more that needs to be commited

MarianoGnu (approved):
Please read previous comment before merging

godot_project/project.godot
HungryProton: My comment was meant to only target line 350 but github included the line above anyway.
Even when manually adding etc2_astc.macos to the project.godot, the export is still blocked
(and removing the s3tc line doesn't change anything either)

Otherwise these are the only .import files affected. I deleted the .godot folder and reopened the project, no other files were modified.

PR #187, opened 2025-05-09
Fix the Alt+H shortcut to unhide hidden objects when Autopose Atoms feature is enabled

MarianoGnu:
- Autopose was consuming every input when ALT was being held down, not allowing ANY shortcut using ALT
- Now the input handler will only block inputs when a candidate is being hovered

Task: BUG: ALT+H shortcut doesn't work to unhide objects (At least on Windows)


HungryProton (changes requested):
The only reason this handler becomes exclusive when holding alt is because the transform gizmo gets in the way, so it was made so pressing alt hides everything else except for the atom candidates.

This PR cancels that behavior. If there is currently no way to block other input handlers but not the shortcuts, I can look into it

PR #192, opened 2025-05-10
Fix the "Show hidden" shortcut not working when atom suggestions are enabled

HungryProton:
Fixes: BUG: ALT+H shortcut doesn't work to unhide objects

This adds an option to let exclusive input handlers to pass the input up to the editor viewport. (Other inputs handlers are still blocked from processing this input)

### DIFF
--- a/godot_project/editor/controls/menu_bar/msep_popup_menu.gd
+++ b/godot_project/editor/controls/menu_bar/msep_popup_menu.gd
@@ -6,7 +6,7 @@ func _init() -> void:
 	MolecularEditorContext.workspace_activated.connect(_on_workspace_activated)
 	about_to_popup.connect(_on_about_to_popup)
 	popup_hide.connect(_on_popup_hide)
-	id_pressed.connect(_on_forward_id_pressed)
+	id_pressed.connect(_on_id_pressed)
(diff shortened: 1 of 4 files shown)


godot_project/editor/controls/menu_bar/msep_popup_menu.gd
```diff
@@ -64,15 +64,5 @@ func _update_menu() -> void:
...
 
-func _on_forward_id_pressed(in_id: int) -> void:
-	var workspace_context: WorkspaceContext = MolecularEditorContext.get_current_workspace_context() as WorkspaceContext
-	if is_instance_valid(workspace_context):
-		if workspace_context.get_editor_viewport().has_exclusive_input_consumer():
-			# block menu activations (assumed from keyboard shortcuts)
-			# when viewport has an exclusive consumer
-			return
-	_on_id_pressed(in_id)
-
```
HungryProton, line 75: This was never called, or rather, it jumped to `_on_id_pressed` directly all the time.
Because the input is set as handled in viewport_input.gd:49, the `id_pressed` event is not triggered when there is an exclusive input handler.

MarianoGnu (approved):

PR #193, opened 2025-05-11
Update the about screen to be legible on both dark and bright backgrounds

HungryProton:
Fixes: BUG - About Panel invisible when in Academic Mode

I tried to keep it as close as possible to the original design, it's the same colors in both cases.
If the contrast isn't high enough, we'll have to either drop the transparency, or make two variants and dynamically switch depending on the background brightness.

Dark:
image

Bright:
image

Before:
image


MarianoGnu (approved):
sheep it 🐑

MarianoGnu:
actually, this might need a followup PR, `InitialInfoScreen` most probably will need similar treatment.
usually this screen appears on top of a brand new workspace, which has always dark backgroud
but if the user opens the application by opening a file from command line or because they asociated the file extension to the application, then the same problem will happen

image

MarianoGnu:
I merged this because it was requested by jonathan, but please check my message

PR #195, opened 2025-05-12
Add a visual representation of simulation boundaries when area is constrained

MarianoGnu:
- Also added a check button to Visibility section in WorkspaceSettings to toggle it on and off

Task: Show Simulation Bounding Cube Option

### DIFF
--- a/godot_project/editor/rendering/rendering.gd
+++ b/godot_project/editor/rendering/rendering.gd
@@ -37,6 +37,8 @@ signal representation_changed(new_representation: Representation)
 @onready var _world_environment: WorldEnvironment = $WorldEnvironment
 @onready var _spring_preview: SpringPreview = $SpringPreview
 @onready var _selection_preview: SelectionPreview = $SelectionPreview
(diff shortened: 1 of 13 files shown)


godot_project/editor/rendering/rendering.gd
```diff
@@ -857,10 +861,40 @@ func spring_preview_hide() -> void:
...
+	_update_simulation_boundaries_visibility()
+
+
+func _update_simulation_boundaries_visibility() -> void:
+	if not is_instance_valid(_workspace_context):
+		_simulation_boundaries_representation.hide()
+		return
+	var workspace: Workspace = _workspace_context.workspace
+	if not is_instance_valid(workspace):
+		push_error("Workspace is not valid")
```
HungryProton, line 885: if the workspace isn't valid, you should probably ensure the boundaries are hidden too

HungryProton (approved):

godot_project/editor/rendering/rendering.gd
MarianoGnu: ah, good point

PR #196, opened 2025-05-13
Support for relaxing martial selection of molecules

MarianoGnu:
- When relaxing partial selection, neighbors unselected atoms will be sent, but be locked in place and passivated

Task: ENHANCEMENT: Support for relaxing partial selection of a molecule

### DIFF
--- a/godot_project/autoloads/openmm/openmm_payload.gd
+++ b/godot_project/autoloads/openmm/openmm_payload.gd
@@ -25,7 +25,7 @@ var topology_bonds: PackedByteArray = []
 var topology_passivated_atoms: PackedByteArray = []
 var atoms_state: PackedByteArray = []
 var passivation_state: PackedByteArray = []
-var request_atom_id_to_structure_and_atom_id_map: Dictionary = {
+var request_atom_id_to_structure_and_atom_id_map: Dictionary[int,PackedInt32Array] = {
 #	request_atom_id: int = [structure_int_guid: int, atom_id: int]
 }
 var raw_initial_positions: PackedVector3Array = [] # Unaltered positions from the source structure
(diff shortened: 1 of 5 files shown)


godot_project/autoloads/openmm/openmm_payload.gd
```diff
@@ -372,6 +432,22 @@ func _passivate_atom(in_structure: AtomicStructure, in_atom_id: int) -> PackedVe
...
+			continue
+		var bond_order: int = in_structure.get_bond(bond_id).z
+		used_valence += bond_order
+	var valence_left: int = data.valence
+	if data.number > 5:
+		valence_left = 8 - valence_left
+	valence_left -= used_valence
+	return valence_left
+
+
```
HungryProton, line 450: Notes for a future PR: 
There is something similar in `add_posing_atom_input_handler.gd`, (mainly, the diff between `get_charge()` and `get_stable_charge()`) but also in `ring_action_add_hydrogens.gd`

It would be great to have a single place where we do these calculations instead of having 3 different implementations for the same thing.

godot_project/editor/controls/dockers/workspace_docker/simulations_docker/simulations_docker_controls/relax_tools_panel.tscn
```diff
@@ -61,13 +61,6 @@ layout_mode = 2
 button_pressed = true
 text = "Treat incomplete valences as if they have been filled by hydrogens"
 
-[node name="LabelSelectOnlyNotice" type="Label" parent="."]
-unique_name_in_owner = true
-layout_mode = 2
-text = "ℹ When relaxing selected molecules, all atoms in the molecule should be selected"
-horizontal_alignment = 1
-autowrap_mode = 2
-
```
HungryProton, line 70: Maybe also rename "Only selected molecules" by "Selection only", ( the checkbox inside `PanelContainer/VBoxContainer` )

HungryProton (approved):

godot_project/autoloads/openmm/openmm_payload.gd
MarianoGnu: Yes, but this code is special, because is (potentially) ignoring some of the bonds because the neighbor atoms are not in selection or are neighbor of selection, this is why we have some duplicated code here

godot_project/editor/controls/dockers/workspace_docker/simulations_docker/simulations_docker_controls/relax_tools_panel.tscn
MarianoGnu: Good thinking

PR #198, opened 2025-05-15
Initial implementation of Particle Emitters (WIP)

MarianoGnu:
- Hidden under feature flags (use F12 to toggle it)
- Created resources with the definition of the class, making possible to save it to disk
- Created a "Create Panel" that can edit what properties will have a mewly created Particle Emitter
- Known issues: "Create Panel" cannot yet assign the template of the molecule that will be instantiated

Task: Molecular Emitter


godot_project/project_workspace/workspace_context/create_object_parameters.gd
```diff
@@ -22,11 +22,12 @@ signal validate_bonds_requested(selection_only: bool)
...
-	CREATE_SHAPES,
-	CREATE_FRAGMENT,
-	CREATE_VIRTUAL_MOTORS,
-	CREATE_ANCHORS_AND_SPRINGS,
+	CREATE_ATOMS_AND_BONDS     = 0,
+	CREATE_SHAPES              = 1,
+	CREATE_FRAGMENT            = 2,
+	CREATE_VIRTUAL_MOTORS      = 3,
+	CREATE_PARTICLE_EMITTERS   = 4,
+	CREATE_ANCHORS_AND_SPRINGS = 5,
```
HungryProton, line 30: Is this stored on disk at any point?
If not we probably don't need to enforce a value for each.

HungryProton (approved):

PR #199, opened 2025-05-15
Fix warnings when launching the project

MarianoGnu:
No Task

This node added to EditorViewport will prevent the annoying "2D MSAA is enabled while there is no 2D content" warning
image

### DIFF
--- a/godot_project/editor/rendering/reference_shape_renderer/assets/reference_shape_renderer_material.tres
+++ b/godot_project/editor/rendering/reference_shape_renderer/assets/reference_shape_renderer_material.tres
@@ -1,7 +1,9 @@
 [gd_resource type="ShaderMaterial" load_steps=2 format=3 uid="uid://wvju7e8kntu2"]
 
-[ext_resource type="Shader" path="res://editor/rendering/reference_shape_renderer/reference_shape_shader.gdshader" id="1_wem1y"]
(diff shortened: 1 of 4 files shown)


godot_project/editor/rendering/reference_shape_renderer/assets/reference_shape_renderer_material.tres
```diff
@@ -1,7 +1,9 @@
 [gd_resource type="ShaderMaterial" load_steps=2 format=3 uid="uid://wvju7e8kntu2"]
 
-[ext_resource type="Shader" path="res://editor/rendering/reference_shape_renderer/reference_shape_shader.gdshader" id="1_wem1y"]
+[ext_resource type="Shader" uid="uid://cx6ygy56gp742" path="res://editor/rendering/reference_shape_renderer/reference_shape_shader.gdshader" id="1_wem1y"]
 
 [resource]
 render_priority = 0
 shader = ExtResource("1_wem1y")
+shader_parameter/default_wire_width = 0.8
+shader_parameter/hovered_wire_width = 2.4
```
MarianoGnu, line 9: Godot insisted to modify this file every time i opened it.
Besides adding the shader uid, these properties values are the same as the default values on the shader, seems like there where was a change in Godot to include the values in the material even if it is the default, (which concerns me because if i modify the default value from the shader it could not affect the material, specially if i change it from an external editor)

HungryProton (approved):

PR #200, opened 2025-05-17
Add ParticleEmitterRenderer to Rendering API

MarianoGnu:
- There's yet no way to actually create one
- Simplified some repetitive code in Rendering class

DEPENDS ON #198

Task: Molecular Emitter


godot_project/editor/rendering/rendering.gd
```diff
@@ -220,6 +224,13 @@ func is_renderer_for_motor_built(in_motor: NanoVirtualMotor) -> bool:
 	return true
 
 
+func is_renderer_for_particle_emitter_build(in_emitter: NanoParticleEmitter) -> bool:
```
HungryProton, line 227: nit: this should be `built` with a t at the end, not a d

HungryProton (approved):

godot_project/editor/rendering/rendering.gd
MarianoGnu: ah, will fix it in a latter PR, i dont want to rebase a lot of work only for this

PR #202, opened 2025-05-19
Fix atom overlaps detection accross different groups

HungryProton:
Fixes: BUG - Validation Succeeds when it should fail

The hash grid was recreated for each group, instead of using a single one for the whole selection.

This also changes some values in the fix_overlapping_atoms feature method, since right now, that button doesn't move the atoms far enough to fix the overlap.

### DIFF
--- a/godot_project/editor/controls/dockers/workspace_docker/simulations_docker/simulations_docker_controls/atomic_structure_model_validator.gd
+++ b/godot_project/editor/controls/dockers/workspace_docker/simulations_docker/simulations_docker_controls/atomic_structure_model_validator.gd
@@ -42,12 +42,13 @@ func _validate_bonds_in_thread(
 		out_promise: Promise,
 		in_selection_only: bool) -> void:
 	var validation_results: Array[Metadata] = []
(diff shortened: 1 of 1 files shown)


MarianoGnu (changes requested):
Here is a problem i was expecting, which is not in the detection, is in the UX of the validation

image

Because we cannot select single atoms of different groups (to do that we have to change the currently edited group) i tough we need to create new warning messages and showing it in pairs
```
- 1 Atom(s) of group WWWW are overlapping with group YYYY
- 2 Atom(s) of group YYYYYY are overlaping with group WWWWW
```

Then when selecting each warning that would make each group editable and selecting overlapping atoms

HungryProton:
Updated the PR to split the alerts by group
image

MarianoGnu:
Detection works very good, the fix overlap tho is not that great, we could iterate in a different PR, making move in `-atom.direction_to(centerof_overlapping_atoms)` direction instead of random

https://github.com/user-attachments/assets/ef3e2ad4-bbc3-4dc5-b7b1-47a66a7c5223

MarianoGnu (approved):

PR #204, opened 2025-05-19
Add input handler to create a particle emitter, also some fixes to rendering and transform handler

MarianoGnu:
- Transfirm input handler was not handlind particle emitters
- StructureContext was not handlind selection of Particle Emitters
- The 3D model of ParticleEmitter didn't have an outline mesh (normals inverted with negative X UV coordinates)
- Known issues: once unselected cannot select emitters again, not supported by CollisionEngine/Space and SelectionInput

DEPENDS ON #198, #200, and #203

Task: Molecular Emitter


MarianoGnu:
Specs changed

PR #205, opened 2025-05-20
Added support for Selection and BoxSelection for Particle Emitters

MarianoGnu:
- Added particle emitters to collision engine, with a rational offset respective to it's 3D model
- Added particle emitters to MultiStructureHitResult processing
- handle particle emitters in Selection (click) and BoxSelection
- Some fixes to 3D model: replaced the material of the handle for a more readable one, and unwraped it's UV's (handle has no selection outline on purpose at the moment)

DEPENDS ON https://github.com/MSEP-one/msep.one/pull/198, https://github.com/MSEP-one/msep.one/pull/200, https://github.com/MSEP-one/msep.one/pull/203, and #204

Task: Molecular Emitter

PR #206, opened 2025-05-20
Added support for editing(+Undo/Redo) properties of selected ParticleEmitter

MarianoGnu:
- (Still lacks capability to set molecule template to spawn)

DEPENDS ON #198, #200, #203, #204, and #205


MarianoGnu:
Specs changed

PR #207, opened 2025-05-21
Add support for setting ParticleEmitter molecule template from Small Molecules Library

MarianoGnu:
DEPENDS ON #200, #203, #204, #205, and #206

Task: Molecular Emitter


MarianoGnu:
Specs changed

PR #208, opened 2025-05-21
Add support to set Particle Emitter's Molecule Template from selected atoms and bonds

MarianoGnu:
- Also allows to clear the template when nothing is selected
- When assigning a molecule template to ParticleEmitter ensure is centered in the origin of coordinates
DEPENDS ON
#200
#203
#204
#205
#206
#207
Task: Molecular Emitter


MarianoGnu:
Specs changed

PR #215, opened 2025-05-28
Fix alerts panel issues

HungryProton:
Fixes: Undo/Redo won't update Alert Panel messages about fixed issues

Also fixes:
+ After creating a group & atoms, and after running a validation about these atoms, undoing everything still shows the alert: clicking it triggers an assert.
+ My laptop doesn't display the unicode glyphs sometimes (despite the font being there, not sure why that happens). The alerts now uses actual icons instead.

---

Alerts panel now registers as a snapshotable system. The internal tree items are no longer exposed, but it generates and return an ID for each alert. You can mark these alerts as fixed or invalid by using this ID.

image

The Fixed and Invalid filter buttons only appear if relevant alerts are present. If not, they're hidden to limit clutter.

image


MarianoGnu:
> The Fixed and Invalid filter buttons only appear if relevant alerts are present. If not, they're hidden to limit clutter.

To consider: Should **Warning** and **Error** have the same treatment?

godot_project/editor/controls/dockers/workspace_docker/simulations_docker/simulations_docker_controls/atomic_structure_model_validator.gd
```diff
@@ -13,16 +13,9 @@ const MAX_SHAKE_ITERATIONS: int = 5
...
-var _overlaps: Dictionary = {
-#	overlap_data<OverlapData> = tree_item<TreeItem>
-}
-var _tree_items: Dictionary = {
-#	tree_item<TreeItem> = true<bool>
-}
-
-
+var _overlaps: Dictionary[OverlapData, int] = {} # Overlap data : Alert ID
+var _data: Dictionary[Metadata, int] = {} # Data : Alert ID
```
MarianoGnu, line 17: nit: i am not super convinced with the name of this variable, the "values" contained in the dictionary is the `_tree_items` IDs as advertised before, Data is the key

HungryProton:
> > The Fixed and Invalid filter buttons only appear if relevant alerts are present. If not, they're hidden to limit clutter.
> 
> To consider: Should **Warning** and **Error** have the same treatment?

I think it's fine, but be could do it. However if there's only one filter button visible, maybe it should be disabled, otherwise if the user click it, they just end up with an empty alert panel, and I find it confusing.

godot_project/editor/controls/dockers/workspace_docker/simulations_docker/simulations_docker_controls/atomic_structure_model_validator.gd
```diff
@@ -96,29 +89,28 @@ func validate_atomic_model(in_selection_only: bool) -> void:
...
-		_tree_items[item] = true
-		item.set_metadata(0, metadata)
+			alert_id = _workspace_context.push_error_alert(metadata.text, _on_alert_selected, _on_alert_selected.bind(true))
+		_data[metadata] = alert_id
 		if metadata is OverlapData:
-			_overlaps[metadata] = item
+			_overlaps[metadata as OverlapData] = alert_id
 	
 	validation_finished.emit(not _overlaps.is_empty())
+	_workspace_context.snapshot_moment("Validate model")
```
MarianoGnu: This is needed only to have and before and after in the snapshots, correct?

godot_project/editor/controls/dockers/workspace_docker/simulations_docker/simulations_docker_controls/validate_bonds_panel.gd
```diff
@@ -119,7 +119,7 @@ func _on_only_selected_checkbox_toggled(_enabled: bool) -> void:
 
 
 func _on_invisible_selection_label_meta_clicked(_meta: Variant) -> void:
-	var selected_alert: TreeItem = _workspace_context.get_alert_selected()
+	var selected_alert: int = _workspace_context.get_alert_selected()
 	if selected_alert:
```
MarianoGnu, line 123: ⚠️  judging by [this line](https://github.com/MSEP-one/msep.one/blob/36e4b9c74b85bc7033fdbb2bc8659433d5f8a285/godot_project/editor/controls/dockers/workspace_docker/simulations_docker/simulations_docker_controls/validate_bonds_panel.gd#L72), _workspace_context.get_alert_selected() will return `-1` if there's nothing selected, meaning this if condition is wrong because `-1` wil evaluate to `true`, ~~and also the first item in the tree with id `0` will be ignored~~ (this assumption was wrong, the lowest id you get is 1, not 0)

godot_project/editor/controls/workspace_view/controls/alerts_panel/alerts_panel.gd
```diff
@@ -1,89 +1,159 @@
 class_name AlertsPanel extends PanelContainer
 
+
+const ICON_WARNING := preload("res://editor/icons/icon_warning_x16.svg")
+const ICON_ERROR := preload("res://editor/icons/icon_error_x16.svg")
+const ICON_FIXED := preload("res://editor/icons/icon_valid_x16.svg")
+const ICON_DELETED: Texture2D = preload("res://editor/controls/menu_bar/menu_edit/icons/icon_delete.svg")
```
MarianoGnu: nit: This is matter of taste, but this could be compressed in a single constant:
```
const ICONS: Dictionary[AlertLevel, Texture2D] = {
    AlertLevel.WARNING = preload("res://editor/icons/icon_warning_x16.svg"),
    AlertLevel.ERROR = preload("res://editor/icons/icon_error_x16.svg"),
    AlertLevel.FIXED = preload("res://editor/icons/icon_valid_x16.svg"),
    AlertLevel.DELETED = preload("res://editor/controls/menu_bar/menu_edit/icons/icon_delete.svg"),
    AlertLevel.INVALID = null,
}
```
simplifying the code where is used
`_tree_item.set_icon(0, ICONS[level])`

godot_project/editor/controls/workspace_view/controls/alerts_panel/alerts_panel.gd
```diff
@@ -1,89 +1,159 @@
...
 
 
-## Returns the selected TreeItem, if any
-func get_alert_selected() -> TreeItem:
-	return _tree.get_selected()
+## Returns the selected alert id, if any
+func get_alert_selected() -> int:
+	var selected: TreeItem = _tree.get_selected()
+	if not selected:
+		return -1
```
MarianoGnu: ⚠️  in a previous comment i mentioned this function returns `-1` wich evaluates to `true`, however seeing `_highest_id` is set to `0` on clear, the lowest id we get is `1`
This means you could return 0 here instead, and fix usages of `get_alert_selected()`
I am fine with any of both ways

MarianoGnu (changes requested):
Awesome work, please address the ⚠️ comments

godot_project/editor/controls/dockers/workspace_docker/simulations_docker/simulations_docker_controls/atomic_structure_model_validator.gd
HungryProton: Yes. This one is optional, I was testing with a without to see which one was the less confusing but I think we should keep the current behavior for now and check with Jonathan later if it needs to be changed.

godot_project/editor/controls/dockers/workspace_docker/simulations_docker/simulations_docker_controls/validate_bonds_panel.gd
HungryProton: Fixed :+1:

godot_project/editor/controls/workspace_view/controls/alerts_panel/alerts_panel.gd
HungryProton: Thanks for the suggestion!

godot_project/editor/controls/workspace_view/controls/alerts_panel/alerts_panel.gd
HungryProton: fixed

MarianoGnu (approved):

PR #217, opened 2025-05-29
Fix loading files containing a particle emitter

HungryProton:
(No card)


MarianoGnu (approved):
This feels like it should be done in the initialization of the NanoStructure, not here

I think i remember a `_post_init` method used for this purpose?

HungryProton:
`_post_init` happens too early, the template structure doesn't have a reference to its related workspace, so it can't get the representation settings. I'm not even sure it ever gets a reference to a workspace later on since this is never really owned by the workspace.

MarianoGnu:
> `_post_init` happens too early, the template structure doesn't have a reference to its related workspace, so it can't get the representation settings. I'm not even sure it ever gets a reference to a workspace later on since this is never really owned by the workspace.

can you override `set_representation_settings` to forward the settings to the template? i am not sure if it is called when the variable is assigned from the tres file...

PR #220, opened 2025-05-30
Ring action for particle emitters

HungryProton:
+ Add a ring action and side menu item for the particle emitters. (Only if the feature flag is enabled)
+ Fix an issue causing the side menu to display outdated items after disabling a feature flag.
  - Outdated items are still there, but disabled.
  - Before, you could click on them and cause an out of bounds error.
+ The quick search panel was only updated once on load. It now regenerate the list of actions if the feature flags were changed since the last time the panel was opened.

(The icon will need to be replaced at some point)


godot_project/editor/controls/menu_bar/visual_menu_content.gd
```diff
@@ -95,6 +95,7 @@ func _update_menu_bar_path(in_menu: PopupMenu, in_path: String) -> void:
 	var category: VisualMenuCategory = candidates.front()
 	var root_menu_category: VisualMenuCategory = category
 	in_menu.about_to_popup.emit() # Force update
+	category.disable_all_items()
```
MarianoGnu, line 101: can we add a comment explaining why we call this? i think i see a point, but i am not sure why, maybe because some paths could simply be missing?

godot_project/editor/controls/quick_search_dialog/quick_search_dialog.gd
```diff
@@ -269,3 +270,8 @@ func _on_tree_item_activated() -> void:
 		_ring_menu.show_in_desired_position(desired_position, fit_in_rect)
 	
 	hide()
+
+
+func _on_feature_flag_toggled(_path: String, _new_value: bool) -> void:
+	_action_items.clear()
+	_menu_items.clear()
```
MarianoGnu, line 282: clearing the items will automatically cause them to repopulate on about_to_show? can we add a comment to explain the logic of why we (only) clear contents here?

MarianoGnu (approved):
Left some comments, address or answer, they are not important

godot_project/editor/controls/menu_bar/visual_menu_content.gd
HungryProton: If we turn off a feature flag, its corresponding entry in the drop down menu will disappear, but not the one from the hamburger menu.
So this method just disables everything at the beginning, and only the valid items will be re enabled in the loop below.
Ideally the items should also be removed but that was the simpler solution.

godot_project/editor/controls/quick_search_dialog/quick_search_dialog.gd
HungryProton: Yes. 
For example, if the user toggles 3 features flags and then opens the quick search menu, the list of actions is rebuilt a single time when the menu is opened, instead of 3 times (1 for each feature flag toggled)

PR #221, opened 2025-05-30
Add support for copy & paste emitters

HungryProton:
Turns out the emitters don't do anything out of the ordinary and can just reuse the existing code paths for copy paste.

### DIFF
--- a/godot_project/autoloads/nano_editor_clipboard/nano_editor_clipboard.gd
+++ b/godot_project/autoloads/nano_editor_clipboard/nano_editor_clipboard.gd
@@ -47,6 +47,8 @@ func copy(in_workspace_context: WorkspaceContext) -> void:
 				_copy_motor(structure_context, nano_structure, new_content)
 			&"AnchorPoint":
 				_copy_anchor(structure_context, nano_structure, new_content)
+			&"ParticleEmitter":
+				_copy_particle_emitter(structure_context, nano_structure, new_content)
 			_:
 				push_warning("Nano structure type not implemented for copy")
 	var root_group_id: int = -1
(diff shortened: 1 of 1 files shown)


godot_project/autoloads/nano_editor_clipboard/nano_editor_clipboard.gd
```diff
@@ -243,6 +245,13 @@ func _copy_anchor(
 	_copy_structure(in_structure_context, in_shape, out_content)
 
 
+func _copy_particle_emitter(
+	in_structure_context: StructureContext,
+	in_shape: NanoParticleEmitter,
+	out_content: Array[Dictionary]) -> void:
+	_copy_structure(in_structure_context, in_shape, out_content)
+
+
```
HungryProton: I've added this to match the existing convention, but why are these methods required? Can't we just call `_copy_structure()` directly in the match above?

godot_project/autoloads/nano_editor_clipboard/nano_editor_clipboard.gd
```diff
@@ -243,6 +245,13 @@ func _copy_anchor(
 	_copy_structure(in_structure_context, in_shape, out_content)
 
 
+func _copy_particle_emitter(
+	in_structure_context: StructureContext,
+	in_shape: NanoParticleEmitter,
```
MarianoGnu: nit: rename to `in_emitter`

godot_project/autoloads/nano_editor_clipboard/nano_editor_clipboard.gd
MarianoGnu: well, this distinction only exists to make sure we take in consideration exeptional cases that make occur with new classes, in example if we just leave it as is, and we add some new object that requires special setup, like in example anchors needed repaing of springs, we should account for that.

I was not sure at first glance if calling `_copy_structure()` woould create a duplicate of `_parameters`, and inside of it if it create a duplicate of the `template` object
Seems it works and i was concerned for nothing

MarianoGnu (approved):
sheep it

PR #222, opened 2025-05-31
Fix particle emitters structure preview

HungryProton:
+ Updates the preview position during undo redo
+ Fix an issue where all the previews disappear when changing the `Create object mode` to any option.

### DIFF
--- a/godot_project/editor/rendering/structure_preview/structure_preview.gd
+++ b/godot_project/editor/rendering/structure_preview/structure_preview.gd
@@ -18,17 +18,15 @@ var _preview_workspace: Workspace
 var _preview_structure_context: StructureContext
 var _enabled_on_preview_viewport: bool = false
 
+
 func _ready() -> void:
 	hide()
-
 	_ready_deferred.call_deferred()
 
 
 func _ready_deferred() -> void:
 	var workspace_context: WorkspaceContext = MolecularEditorContext.get_current_workspace_context() as WorkspaceContext
 	assert(workspace_context)
-	workspace_context.started_creating_object.connect(_on_workspace_context_started_creating_object.bind(workspace_context))
-	workspace_context.aborted_creating_object.connect(_on_workspace_context_aborted_creating_object)
 	var representation_settings: RepresentationSettings = workspace_context.workspace.representation_settings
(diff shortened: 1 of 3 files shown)


godot_project/editor/rendering/structure_preview/structure_preview.gd
```diff
@@ -92,14 +90,3 @@ func set_transparency(in_transparency: float) -> void:
...
-func _on_workspace_context_started_creating_object(workspace_context: WorkspaceContext) -> void:
-	var peek_new_object: Callable = func(in_context: StructureContext) -> bool:
-		set_structure(in_context.nano_structure)
-		return true
-	workspace_context.peek_context_of_object_being_created(peek_new_object)
-
-
-func _on_workspace_context_aborted_creating_object() -> void:
-	if is_instance_valid(_atomic_structure_renderer):
-		_atomic_structure_renderer.queue_free()
```
HungryProton, line 105: I don't remember why this was needed initially, but it doesn't seem to be used anymore and it's causing the second bug mentioned in this PR.

godot_project/project_workspace/structs/nano_particle_emitter.gd
```diff
@@ -238,3 +238,4 @@ func apply_state_snapshot(in_state_snapshot: Dictionary) -> void:
 	if _parameters == null:
 		_parameters = NanoParticleEmitterParameters.new()
 	_parameters.apply_state_snapshot(in_state_snapshot["_parameters_snapshot"])
+	transform_changed.emit(_transform)
```
MarianoGnu: This goes against the design rules of the snapshot system, the full idea is we should not need to emit any change from here.
The actual mistake seems to be happening in `particle_emitter_renderer.gd` on `apply_state_snapshot`, it is setting `global_transform = in_snapshot["global_transform"]` but is not updating the position of the template preview

MarianoGnu (changes requested):

godot_project/project_workspace/structs/nano_particle_emitter.gd
HungryProton: Fixed!

MarianoGnu (approved):

PR #223, opened 2025-06-02
Calculate offset of multiple instances and show this offset in the emitter renderer preview

MarianoGnu:
Task: Molecular Emitter

https://github.com/user-attachments/assets/e03249b2-c934-433b-846d-1ca290466d42

https://github.com/user-attachments/assets/84e078bf-7ed5-4143-a922-0b7ed25a003f

### DIFF
--- a/godot_project/editor/rendering/particle_emitter_renderer/particle_emitter_renderer.gd
+++ b/godot_project/editor/rendering/particle_emitter_renderer/particle_emitter_renderer.gd
@@ -3,7 +3,7 @@ class_name ParticleEmitterRenderer extends Node3D
 
 var _emitter_id: int
 var _workspace_context: WorkspaceContext
-var _structure_preview: StructurePreview
+var _structure_previews: Array[StructurePreview]
 
 
 var _materials: Array[ShaderMaterial]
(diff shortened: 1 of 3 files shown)


godot_project/editor/rendering/particle_emitter_renderer/particle_emitter_renderer.gd
```diff
@@ -81,21 +81,53 @@ func transform_by_external_transform(in_selection_initial_pos: Vector3, in_initi
...
+			_structure_previews.push_back(instance)
 
 
 func _on_emitter_transform_changed(in_transform: Transform3D) -> void:
 	global_transform = in_transform
-	_structure_preview.global_position = in_transform.origin
+	var emitter: NanoParticleEmitter = _workspace_context.workspace.get_structure_by_int_guid(_emitter_id)
+	for i: int in _structure_previews.size():
+		var preview: StructurePreview = _structure_previews[i]
+		preview.global_position = in_transform.origin + emitter.calculate_instance_offset(i)
```
HungryProton, line 115: Suggestion: Since `calculate_instance_offset` will always return the same result, it could be cached in each `StructurePreview` directly, and updated when the preview count changes. This way there's no need to redo all the expensive calculations each frame while the emitter is moving.

HungryProton (approved):

godot_project/editor/rendering/particle_emitter_renderer/particle_emitter_renderer.gd
MarianoGnu: that's a good idea, lets implement it in a following PR

PR #227, opened 2025-06-04
Make OpenMM server immediately stop simulation on case of a failure

MarianoGnu:
No Task


godot_project/python/scripts/openmm_server.py
```diff
@@ -861,7 +861,7 @@ def __init__(self, simulation_id, publish_scoket, publish_socket_lock, reportInt
 
 	def describeNextReport(self, simulation):
 		if self._has_error:
-			return 0
+			return (0, False, False, False, False, None)
```
MarianoGnu, line 864: this `return 0` was causing a silent error that was never informed back to MSEP

HungryProton (approved):

PR #230, opened 2025-06-05
Fix hiding non virtual objects

MarianoGnu:
No Task


godot_project/utils/workspace_utils.gd
```diff
@@ -359,7 +359,7 @@ static func hide_selected_objects(out_workspace_context: WorkspaceContext) -> vo
 				atomic_structure.set_springs_visibility(springs_to_hide, false)
 			structure_context.set_anchor_selected(false)
 			structure_context.nano_structure.set_visible(false)
-		elif structure_context.is_virtual_object_selected():
+		elif structure_context.nano_structure.is_virtual_object() and structure_context.is_virtual_object_selected():
```
MarianoGnu, line 362: Turns out when i changed this previously, i tested hiding virtual objects and it worked as expected, but `structure_context.is_virtual_object_selected()` asserts if structure_context belongs to a non virtual object

HungryProton (approved):

PR #235, opened 2025-06-05
Fix detection of atoms overlaping inside the same group

MarianoGnu:
Task: (Flollow up to) Undo/Redo won't update Alert Panel messages about fixed issues

### DIFF
--- a/godot_project/editor/controls/dockers/workspace_docker/simulations_docker/simulations_docker_controls/atomic_structure_model_validator.gd
+++ b/godot_project/editor/controls/dockers/workspace_docker/simulations_docker/simulations_docker_controls/atomic_structure_model_validator.gd
@@ -563,14 +563,16 @@ class SpatialHashGridOverlaps extends SpatialHashGrid:
 							overlapping_atoms[other_atom.structure_context] = []
 						overlapping_atoms[other_atom.structure_context].push_back(other_atom)
 						visited[j] = true
-				if overlapping_atoms.size() <= 1:
-					continue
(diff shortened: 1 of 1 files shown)


godot_project/editor/controls/dockers/workspace_docker/simulations_docker/simulations_docker_controls/atomic_structure_model_validator.gd
```diff
@@ -563,11 +563,11 @@ class SpatialHashGridOverlaps extends SpatialHashGrid:
 							overlapping_atoms[other_atom.structure_context] = []
 						overlapping_atoms[other_atom.structure_context].push_back(other_atom)
 						visited[j] = true
-				if overlapping_atoms.size() <= 1:
-					continue
```
HungryProton, line 567: This condition is there to prevent false positives, removing it causes a lot of overlap alerts about atoms who are not overlapping with anything 
image

HungryProton (changes requested):

HungryProton (approved):

godot_project/editor/controls/dockers/workspace_docker/simulations_docker/simulations_docker_controls/atomic_structure_model_validator.gd
MarianoGnu: Is not the amount of groups what matters, but the amount of atoms, i updated the PR to add this check in the right place

PR #238, opened 2025-06-06
Disable constraints when using Particle Emitters

MarianoGnu:
- They cause NaN positions and velocities on particles

Task: Molecular Emitter

### DIFF
--- a/godot_project/python/scripts/openmm_server.py
+++ b/godot_project/python/scripts/openmm_server.py
@@ -951,9 +951,11 @@ class ImportFileResponse():
 		self.bonds = bonds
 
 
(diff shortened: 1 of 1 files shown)


HungryProton (approved):
We should probably warn the user that simulations with and without emitters will behave differently.

MarianoGnu:
I asked David Lexter for input about this solution, i want to hear his opinion, or any other expert's

DavidLister:
This looks good on the technical standpoint, though I think interaction via VdW and partial charges (for molecules) might need a separate treatment. As mentioned elsewhere, I think setting the mass of the hidden atoms to the highest possible value is a great workaround. This means they can absorb a lot of kinetic energy with very small change to their velocity, so even if some interactions remain, negative effects should be minimized.

PR #242, opened 2025-06-09
Hide the 'Molecules per emission' count option of particle emitters under a feature flag

MarianoGnu:
Tasks:
- BUG - Emitter Creates Multiple Molecules, even when "Number of Molecules" is set to 1
- BUG - All pre-created molecules shown as visible when setting number of molecules to larger number


HungryProton (approved):
NIT: there's only one m in 'emit' but I think this feature won't last long anyway

PR #245, opened 2025-06-10
Ensure autoposing candidates are updated when undo history changes

MarianoGnu:
Task: BUG: Can undo creation of atoms leaving autoposing candidates behind

### DIFF
--- a/godot_project/editor/input_handlers/molecular_structures/add_posing_atom_input_handler.gd
+++ b/godot_project/editor/input_handlers/molecular_structures/add_posing_atom_input_handler.gd
@@ -172,6 +172,10 @@ func set_preview_position(_in_position: Vector3) -> void:
 func _update_candidates_if_needed() -> void:
 	if not _candidates_dirty or not _should_show():
 		return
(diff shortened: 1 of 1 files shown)


HungryProton (commented):
I'm not a big fan of adding a new parameter just for this fix. 
A simpler approach would be to  modify `update_candidates_if_needed` like this:

```
if not should_show():
	_hide_preview()
        return
if not _candidates_dirty:
		return
```

But I can't reproduce the bug on my end so I'm not 100% sure this works

MarianoGnu:
problem is `update_candidates_if_needed()` is called from a lot of places, but we only care to force-update when making an undo operation, i also considered having a `force` argument in `update_candidates_if_needed()`
maybe i should split it in two methods and call `_update_candidates()` directly

MarianoGnu:
for more elaboration, with this change, flow of execution when making an undo would be the same as it was before #233
This change made the code stop updating candidates on undo because Ctrl is held down

HungryProton (approved):

PR #246, opened 2025-06-12
Fix a crash when changing an emitter's parameters during a simulation

HungryProton:
Fixes: "CRASH: Application can crash if making changes to particle emitter while simulation is ongoing"

The issue was that snapshot_moment() is async when a simulation is running with emitters (because of the await process in the loop).
This means we were still receiving new frames from openmm when the simulation should have been stopped already.


godot_project/project_workspace/structs/nano_particle_emitter.gd
```diff
@@ -334,7 +334,7 @@ func apply_state_snapshot(in_state_snapshot: Dictionary, in_with_instances: bool
 		_instances_group = group_structure_context.nano_structure as AtomicStructure
 		_instances_group.apply_state_snapshot(in_state_snapshot["_instances_group_state"])
 		group_structure_context.get_collision_engine().rebuild(group_structure_context)
-		var workspace_cotext: WorkspaceContext = MolecularEditorContext.get_current_workspace_context()
-		var rendering: Rendering = workspace_cotext.get_rendering()
```
HungryProton, line 338: This is just fixing a typo, unrelated to the actual fix

MarianoGnu (approved):

PR #255, opened 2025-06-17
Prevent save_as to cause duplicate workspace resources from the same file.

HungryProton:
Fixes: "BUG - Loading files"

Repro steps:
+ Open a project "A"
+ Modify the project and save it as "B"
+ Re-open project "A"
+ It will load "Project B"

This happens because when project A is loaded, the resource cache associates path A with project A.
When saving project A as project B, path B is associated with project B in the cache, but path A is still linked to the same resource (now project B)

As far as I know, there's no way to explicitly tell the `ResourceLoader` to clear a specific resource from the cache.
Calling `take_over_path("")` on the resource **does not** clear it from the cache either.

So this PR bypasses the cache completely when loading a workspace.

### DIFF
--- a/godot_project/autoloads/MolecularEditorContext.gd
+++ b/godot_project/autoloads/MolecularEditorContext.gd
@@ -92,25 +92,26 @@ func find_workspace_possessing_structure(in_structure_to_find: NanoStructure) ->
 
 ## Soft loading a workspace will load it from disk but will not immediately activate it
 func soft_load_workspace(in_path: String) -> Workspace:
-	var workspace: Workspace = load(in_path) as Workspace
+	# Ignore the cache when trying to load the workspace to avoid issues with "Save as"
+	# If the workspace at `in_path` is already open, calling load again will
+	# clear its `resource_path`. This needs to be restored after loading. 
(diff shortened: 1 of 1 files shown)


godot_project/autoloads/MolecularEditorContext.gd
```diff
@@ -92,25 +92,26 @@ func find_workspace_possessing_structure(in_structure_to_find: NanoStructure) ->
...
+	if is_instance_valid(open_workspace):
+		# File is already open in another tab, suggest a new path to prevent
+		# overriding the original.
+		workspace.resource_path = ""
+		workspace.suggested_path = _get_suggested_path(in_path)
+		open_workspace.take_over_path(in_path)
 	
-	var workspace_copy: Workspace = ResourceLoader.load(in_path, "", ResourceLoader.CACHE_MODE_IGNORE)
-	# ResourceLoader.CACHE_MODE_IGNORE ignores cache, but also removes from the cache the object that was
-	# asociated to in_path. This workaround should reasociate the original workspace to the path
```
HungryProton, line 106: This comment doesn't match what I saw when fixing this bug. Calling load this way does clear the resource_path on the resource, but not the resource loader cache.

MarianoGnu (approved):

PR #260, opened 2025-06-19
Abort simulation early if particles are moving impossibly fast

HungryProton:
Mitigates "BUG - Strange Behavior When scrolling through failed simulation" and other similar issues involving a failed, exploded simulation.

Detecting high velocities means we can stop the simulation before it becomes an exploded mess, instead of only checking if the particle distance from the origin is greater than 1e6 units.

~~The speed threshold is set at 1e15 nm/ps to avoid false positives, but I suspect there's something wrong with the unit conversion, or the speeds reported by openff are not actually nm/ps.
(Lower than 1e10 and some simulations that used to work fine are reported as invalid due to high speeds)~~
I wasn't testing that properly, the threshold was lowered to something more reasonable.

### DIFF
--- a/godot_project/python/scripts/openmm_server.py
+++ b/godot_project/python/scripts/openmm_server.py
@@ -939,34 +939,44 @@ class ZmqPublishReporter(object):
 		# - Whether the next report will need forces.
 		# - Whether the next report will need energies.
 		# - Whether the positions should be wrapped to the periodic box. If None, it will automatically decide whether to wrap positions based on whether the System uses periodic boundary conditions.
-		return (steps, True, False, False, False, None)
+		return (steps, True, True, False, False, None)
 
 	def report(self, simulation, state):
 		simulation.frame += 1
 		#time_in_femtoseconds: float = (state.getTime() / femtosecond)
 		#time_buffer:bytes = struct.pack("d", time_in_femtoseconds)
 		time_buffer:bytes = struct.pack("d", simulation.frame)
-		openff_positions: list[Vec3] = state.getPositions()
+		
+		MAXIMUM_SPEED_SQUARED = 1e12
+		for vel in state.getVelocities():
+			v = vel._value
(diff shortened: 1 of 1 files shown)


godot_project/python/scripts/openmm_server.py
```diff
@@ -939,19 +939,22 @@ def describeNextReport(self, simulation):
...
 		time_buffer:bytes = struct.pack("d", simulation.frame)
 		openff_positions: list[Vec3] = state.getPositions()
+		openff_velocities: list[Vec3] = state.getVelocities()
 		positions: list[Vec3] = []
+		velocities: list[Vec3] = []
 		payload_to_openff_atom = simulation.payload_to_openff_atom
 	
 		for i in range(simulation.atoms_count):
 			positions.append(openff_positions[payload_to_openff_atom[i]])
+			velocities.append(openff_velocities[payload_to_openff_atom[i]])
```
MarianoGnu: this is a little unnecesary, you could just use openff_velocities in the check in line 978 and save some memory and cpu cycles

godot_project/python/scripts/openmm_server.py
```diff
@@ -967,6 +970,19 @@ def report(self, simulation, state):
 						self._has_error = True
 						break
 					positions_buffer += struct.pack("d", raw)
+			
+			MAXIMUM_SPEED_SQUARED = 1e30
+			for atom_id, vel in enumerate(velocities):
+				if atom_id >= simulation.atoms_count:
+					break
+				v = vel / nanometer * picosecond
```
MarianoGnu: another way to remove the unit from the vector: `v = vel._value`

godot_project/python/scripts/openmm_server.py
MarianoGnu: you could also make the check right here and avoid the second iteration and enumeration

MarianoGnu (approved):

PR #265, opened 2025-06-20
Dont show full path in dialog when 'Saving file As'

MarianoGnu:
Task: BUG: "Save as" puts the full path as a suggestion for the new file name

### DIFF
--- a/godot_project/editor/MolecularEditor.gd
+++ b/godot_project/editor/MolecularEditor.gd
@@ -139,8 +139,14 @@ func show_open_workspace_dialog() -> void:
 
 func show_save_workspace_dialog(in_workspace: Workspace) -> void:
 	save_file_dialog.set_meta(&"__workspace__", weakref(in_workspace))
-	if not in_workspace.suggested_path.is_empty():
-		save_file_dialog.current_path = in_workspace.suggested_path
+	var path: String = in_workspace.suggested_path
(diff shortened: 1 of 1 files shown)


godot_project/editor/MolecularEditor.gd
```diff
@@ -139,8 +139,14 @@ func show_open_workspace_dialog() -> void:
 
 func show_save_workspace_dialog(in_workspace: Workspace) -> void:
 	save_file_dialog.set_meta(&"__workspace__", weakref(in_workspace))
-	if not in_workspace.suggested_path.is_empty():
-		save_file_dialog.current_path = in_workspace.suggested_path
+	var path: String = in_workspace.suggested_path
+	if path.is_empty():
+		# no suggested dir, fallback to last saved path
+		path = in_workspace.resource_path
+	if not path.is_empty():
```
HungryProton, line 146: this could just be an `else`

HungryProton (approved):

godot_project/editor/MolecularEditor.gd
MarianoGnu: no, because i am overriding the value of path inside the previous if condition

HungryProton: Oh right

PR #279, opened 2025-06-25
Fix the small molecule picker preview with selected items

HungryProton:
Fixes: BUG - Image of selected Small Molecule not visible if not rolled over in 'Create from Small Molecule' feature.


MarianoGnu (changes requested):
`_set_hovered_item(null)` was manually used to clear the texture. There should never be a selected item, because as soon as you click an item this dialog closes 🤔
instead the fix should be set selected item to null on `about_to_popup`

Lets discuss this this before merging

MarianoGnu (approved):

PR #280, opened 2025-06-25
Improve the "Fix overlaps" feature

HungryProton:
Currently, this feature handles overlaps independently from each other, while ignoring the rest of the workspace.

So if two carbons are overlapping, this feature will move them away, potentially inside other atoms that were not part of the overlap.

This PR does two things:
+ All overlapping atoms are relaxed as once during a single pass (instead of multiple pass per overlaps)
+ The relaxation also takes into account the other atoms from the workspace (These atoms won't move, but will influence the relaxation)

### DIFF
--- a/godot_project/editor/controls/dockers/workspace_docker/simulations_docker/simulations_docker_controls/atomic_structure_model_validator.gd
+++ b/godot_project/editor/controls/dockers/workspace_docker/simulations_docker/simulations_docker_controls/atomic_structure_model_validator.gd
@@ -10,6 +10,8 @@ const COLOR_DELETED: Color = Color.WEB_GRAY
 const DELETED_ICON: Texture2D = preload("res://editor/controls/menu_bar/menu_edit/icons/icon_delete.svg")
 const MAX_COVALENT_RADIUS: float = 0.232
 const MAX_SHAKE_ITERATIONS: int = 5
+# Cell size needs be large enough to hold the biggest atom + some margin of error
+const MAX_HASH_GRID_CELL_SIZE: float = MAX_COVALENT_RADIUS * 3.0
 
 
 var _thread: Thread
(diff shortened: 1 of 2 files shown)


godot_project/utils/spatial_hash_grid/spatial_hash_grid.gd
```diff
@@ -40,14 +40,60 @@ func _init(in_cell_size: float) -> void:
...
+		_grid[new_cell_id] = []
+	_grid[new_cell_id].push_back(item)
+
+
+## Returns a flat list of all items in the grid
+func get_all_items() -> Array[Item]:
+	var result: Array[Item] = []
+	for list: Array in _grid.values():
+		for item: Item in list:
+			result.push_back(item)
```
MarianoGnu: nit: instead of
```
		for item: Item in list:
			result.push_back(item)
```
you can do
```
result.append_array(list)
```
Disclaimer: not sure if it would complain of `Array` vs `Array[Item]`, but because _grid is typed you should be able to declare line 93 as
```
for list: Array[Item] in _grid.values():
```

godot_project/utils/spatial_hash_grid/spatial_hash_grid.gd
```diff
@@ -109,11 +155,27 @@ func get_items_closer_than(distance: float) -> Array[Array]:
...
 
 
+func get_items_around(position: Vector3, distance: float = -1.0) -> Array[Item]:
+	assert(distance <= _cell_size, "Max distance can't be larger than the grid's cell_size")
+	if distance <= 0.0:
+		distance = _cell_size
+	var result: Array[Item] = []
+	var distance_sqrd: float = pow(distance, 2.0)
+	var center_cell_id: Vector3 = snapped(position, _snap)
+	var cells_to_scan: Array[Vector3] = get_neighbor_cells(center_cell_id)
```
MarianoGnu, line 164: is it possible to have any neighbor Item when distance cannot be larger than the _cell_size?

godot_project/editor/controls/dockers/workspace_docker/simulations_docker/simulations_docker_controls/atomic_structure_model_validator.gd
```diff
@@ -126,76 +126,117 @@ func has_overlapping_atoms() -> bool:
...
 			continue
-		var nano_structure: NanoStructure = overlap.structure_context.nano_structure
-		if not nano_structure.is_being_edited():
-			nano_structure.start_edit()
-		
-			var original_atoms: PackedInt32Array = nano_structure.get_visible_atoms()
-			var original_positions: PackedVector3Array = PackedVector3Array()
-			for original_atom_id: int in original_atoms:
-				original_positions.push_back(nano_structure.atom_get_position(original_atom_id))
+		var nano_structure: AtomicStructure = overlap.structure_context.nano_structure
```
MarianoGnu, line 147: ❓  unrelated to this PR, but can reference to structure_context and nano_structure break with undo/redo if a group is deleted and then readded with undo? many places in the app only stores structure ID and get it from the _workspace_context for this reason

godot_project/editor/controls/dockers/workspace_docker/simulations_docker/simulations_docker_controls/atomic_structure_model_validator.gd
```diff
@@ -126,76 +126,117 @@ func has_overlapping_atoms() -> bool:
 
 
 func fix_overlapping_atoms() -> void:
+	# Create a hash grid of the whole workspace to avoid moving an overlapping
+	# atom on top of another atom not involved in the original overlap.
+	# This hash grid does not contains the overlapping atoms.
+	var static_atom_grid := SpatialHashGrid.new(MAX_COVALENT_RADIUS * 4.0)
```
MarianoGnu: nit: it took me some time to understand why MAX_COVALENT_RADIUS was multiplied by 4, maybe add a comment, or convert the magic number to a constant?

godot_project/editor/controls/dockers/workspace_docker/simulations_docker/simulations_docker_controls/atomic_structure_model_validator.gd
```diff
@@ -126,76 +126,117 @@ func has_overlapping_atoms() -> bool:
...
 		for atom_id in overlap.atoms_id:
 			if not nano_structure.is_atom_valid(atom_id):
-				# atom was deleted after overlaps was validated
+				# Atom was deleted after the validation
 				continue
 			var atom_position: Vector3 = nano_structure.atom_get_position(atom_id)
-			var random_dir: Vector3 = Vector3(randf() - 0.5, randf() - 0.5, randf() - 0.5).normalized()
-			atoms_positions[atom_id] = atom_position + random_dir * MAX_COVALENT_RADIUS * 0.15
+			overlap_center += atom_position
+			static_atom_grid.remove_item_by_position(atom_position)
```
MarianoGnu, line 160: do you want to remove all atoms at position or only this particular one? should pass `atom_id` as user_data argument?

godot_project/editor/controls/dockers/workspace_docker/simulations_docker/simulations_docker_controls/atomic_structure_model_validator.gd
```diff
@@ -126,76 +126,117 @@ func has_overlapping_atoms() -> bool:
...
-				
-				atoms_positions[atom_id] += dir.normalized() * move_offset
-				atoms_were_moved = true
+		# Move each overlapping atoms away from the center point, by at least
+		# their atomic radius.
+		for atom_id: int in current_overlap_atoms:
+			var item: SpatialHashGrid.Item = current_overlap_atoms[atom_id]
+			var position: Vector3 = item.position
+			if position.is_equal_approx(overlap_center):
+				position += Vector3(randf() - 0.5, randf() - 0.5, randf() - 0.5) * 0.1
```
MarianoGnu: this will create an offset of **at most** 0.05 in each direction, in the case of picking (0.05, 0.05, 0.05) that is a distance of `0.086` from `overlap_center`, is this enough?

the closer randf() is to 0.5, the smaller the distance to overlap_center will be

maybe change `if position.is_equal_approx(overlap_center):` for
```
while position.is_equal_approx(overlap_center):
```

godot_project/editor/controls/dockers/workspace_docker/simulations_docker/simulations_docker_controls/atomic_structure_model_validator.gd
```diff
@@ -126,76 +126,117 @@ func has_overlapping_atoms() -> bool:
...
+			var dir_from_center: Vector3 = overlap_center.direction_to(position)
+			var atomic_number: int = nano_structure.atom_get_atomic_number(atom_id)
+			var element_data: ElementData = PeriodicTable.get_by_atomic_number(atomic_number)
+			var radius: float = element_data.covalent_radius[1]
+			const DISTANCE_MARGIN: float = 1.1
+			var new_position: Vector3 = overlap_center + dir_from_center * radius * DISTANCE_MARGIN
```
MarianoGnu, line 182: you will displace from `overlap_center` wich is shared by all atoms? not from `item.position`? (i understand the difference will only be the distance form the center to the actual atom)

godot_project/editor/controls/dockers/workspace_docker/simulations_docker/simulations_docker_controls/atomic_structure_model_validator.gd
```diff
@@ -126,76 +126,117 @@ func has_overlapping_atoms() -> bool:
...
+			nearby_atoms = overlap_atom_grid.get_items_around(item.position)
+			nearby_atoms.append_array(static_atom_grid.get_items_around(item.position))
+			var closest_atom: SpatialHashGrid.Item
+			var min_distance: float = INF
+			var separation_dir: Vector3
+			for candidate_item in nearby_atoms:
+				if item == candidate_item:
+					continue
+				var diff: Vector3 = item.position - candidate_item.position
+				var distance: float = diff.length_squared()
```
MarianoGnu: nit: `distance_sqrd`

godot_project/editor/controls/dockers/workspace_docker/simulations_docker/simulations_docker_controls/atomic_structure_model_validator.gd
```diff
@@ -126,76 +126,117 @@ func has_overlapping_atoms() -> bool:
...
+	var multiplier: float = 1.0
+	for iteration: int in MAX_SHAKE_ITERATIONS:
+		var atoms_were_moved: bool = false
+		for item: SpatialHashGrid.Item in overlap_atom_grid.get_all_items():
+			# Find the closest atom
+			var nearby_atoms: Array[SpatialHashGrid.Item] = []
+			nearby_atoms = overlap_atom_grid.get_items_around(item.position)
+			nearby_atoms.append_array(static_atom_grid.get_items_around(item.position))
+			var closest_atom: SpatialHashGrid.Item
+			var min_distance: float = INF
```
MarianoGnu: nit: `min_distance_sqrd`

godot_project/editor/controls/dockers/workspace_docker/simulations_docker/simulations_docker_controls/atomic_structure_model_validator.gd
```diff
@@ -126,76 +126,117 @@ func has_overlapping_atoms() -> bool:
...
+			var separation_dir: Vector3
+			for candidate_item in nearby_atoms:
+				if item == candidate_item:
+					continue
+				var diff: Vector3 = item.position - candidate_item.position
+				var distance: float = diff.length_squared()
+				if distance < min_distance:
+					separation_dir = diff
```
MarianoGnu: nit: because this variable claims to be a direction, better to normalize here instead of line 214

MarianoGnu (changes requested):
Looks great! I have some commeents and questions that needs answers before approving tho

godot_project/utils/spatial_hash_grid/spatial_hash_grid.gd
HungryProton: Yes! 

If an item is near the edge of its cell, the neighbors could be in the same cell, or in the cells next to it. So we have to also check the surrounding cells.

If the max distance is limited to the cell size, it guarantees that if there's something, it will be located in that 3x3 grid pattern around its position.

If it's bigger, we need to also check the cells around these 9 cells, potentially more and we loose the speed benefits of doing a hash grid in the first place.

godot_project/utils/spatial_hash_grid/spatial_hash_grid.gd
HungryProton: Seems to work!

godot_project/editor/controls/dockers/workspace_docker/simulations_docker/simulations_docker_controls/atomic_structure_model_validator.gd
HungryProton: In theory, the model validator is not supposed to hold any state, so it probably won't matter, but I can address this in a future PR :+1:

godot_project/editor/controls/dockers/workspace_docker/simulations_docker/simulations_docker_controls/atomic_structure_model_validator.gd
HungryProton: Now that I think about it, x2 would be the minimum value. I'll add a comment

godot_project/editor/controls/dockers/workspace_docker/simulations_docker/simulations_docker_controls/atomic_structure_model_validator.gd
HungryProton: All of them. `static_atom_grid` should only contain the valid atoms not overlapping with anything. Here atom_id is an overlapping atom, so if there's anything else at the same spot, it should be removed too.
I guess it ends up the same in both case anyway, since all the overlapping atoms will go through that loop.

godot_project/editor/controls/dockers/workspace_docker/simulations_docker/simulations_docker_controls/atomic_structure_model_validator.gd
HungryProton: As long as it's not 0, it should be fine, that's just to avoid getting a null separation vector. 
I should make sure the random vector is not 0 however.

godot_project/editor/controls/dockers/workspace_docker/simulations_docker/simulations_docker_controls/atomic_structure_model_validator.gd
HungryProton: Yes, at this point I don't know how much overlap there is with every other atoms. So instead of looping through all of them, I just place each atom away from the overlap center, based on the atom radius.

That should separate them by the smallest distance before the relaxation step

godot_project/editor/controls/dockers/workspace_docker/simulations_docker/simulations_docker_controls/atomic_structure_model_validator.gd
HungryProton: I've renamed the variable instead so the normalized() call only happens once

godot_project/editor/controls/dockers/workspace_docker/simulations_docker/simulations_docker_controls/atomic_structure_model_validator.gd
```diff
@@ -126,76 +128,117 @@ func has_overlapping_atoms() -> bool:
...
-				
-				atoms_positions[atom_id] += dir.normalized() * move_offset
-				atoms_were_moved = true
+		# Move each overlapping atoms away from the center point, by at least
+		# their atomic radius.
+		for atom_id: int in current_overlap_atoms:
+			var item: SpatialHashGrid.Item = current_overlap_atoms[atom_id]
+			var position: Vector3 = item.position
+			var dir_from_center: Vector3 = overlap_center.direction_to(position)
+			while dir_from_center.is_equal_approx(Vector3.ZERO):
```
MarianoGnu, line 175: At first i tought now that you inverted the order of this code `overlap_center.direction_to(position)` would throw an error before this condition is reached, but turns out it doesn't, it just returns `Vector3.ZERO`

MarianoGnu (approved):
Approved

PR #283, opened 2025-06-26
Update documentation

HungryProton:
Fixes: UPDATE Documentation in-app

This PDF (18MB) is now tracked with git LFS


.gitattributes
```diff
@@ -1 +1,2 @@
 *.ogv filter=lfs diff=lfs merge=lfs -text
+godot_project/documentation/msep_documentation.pdf filter=lfs diff=lfs merge=lfs -text
```
MarianoGnu: we could track *.pdf instead

MarianoGnu (approved):
Approved with a minor comment

.gitattributes
HungryProton: Done

PR #284, opened 2025-06-26
Make "Correct Overlapping Atoms" accessible through its own action

HungryProton:
Fix: Add "Correct Overlapping Atoms" feature

This action is exactly the same as going to the validation panel, click "validate", then click "fix overlapping atoms".

(The validation panel is still the same, this PR only adds the ring / menu action)


godot_project/editor/ring_menu_levels/atoms/ring_action_correct_overlapping_atoms.gd
```diff
@@ -0,0 +1,35 @@
...
+
+
+func get_icon() -> RingMenuIcon:
+	return RingMenuSpriteIconScn.instantiate().init(preload("res://editor/icons/icon_fix_overlap.svg"))
+
+
+func _execute_action() -> void:
+	_ring_menu.close()
+	await _model_validator.validate_atomic_model(false)
+	_model_validator.fix_overlapping_atoms()
```
MarianoGnu, line 35: I see a possible followup to this PR:
I think we should first check if there was any overlap, and open the Alerts panel if there was, to show a breakdown of the changes.
I also think we could add an aditional argument to `validate_atomic_model` or call another function `validate_atomic_model_overlaps` to not add any other warning to the alers that is not related to overlaps

image

MarianoGnu (approved):

godot_project/editor/ring_menu_levels/atoms/ring_action_correct_overlapping_atoms.gd
HungryProton: Yes that makes sense, I'll open a card to track this

PR #287, opened 2025-06-27
Add support to cover or fill shapes with small molecules

MarianoGnu:
- The UI can now select individual atoms from periodic tables or small moleculers from our library
- AtomicStructure.get_aabb now has an additional argument to consider the size of the atoms
- When applying molecules to a shape, the longest side of it's AABB is used as a radius or a sphere, then uses the same algorithm for positioning as for atoms

Task: ENHANCEMENT: Support for filling a reference shape with instances of the Small Molecules library besides unbonded atoms

https://github.com/user-attachments/assets/7aba4738-d9f7-43d7-8b29-040bbe8ca4cb

https://github.com/user-attachments/assets/28bfc949-ac41-497a-8d2a-917c99942510

### DIFF
--- a/godot_project/autoloads/PeriodicTable.gd
+++ b/godot_project/autoloads/PeriodicTable.gd
@@ -32,7 +32,7 @@ const NON_METALS = [1, 6, 7, 8, 9, 14, 15, 16, 17]
 var _elements : Dictionary
 var _elements_by_name : Dictionary
 var _elements_by_symbol : Dictionary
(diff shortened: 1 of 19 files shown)


MarianoGnu:
Known Issue: small molecules are not always always centered in their pivor position:

image

MarianoGnu:
> Known Issue: small molecules are not always always centered in their pivor position:

Fixed in 84ac32a

image

HungryProton (approved):

PR #288, opened 2025-06-30
Fix a crash when importing a file from the library during a simulation

HungryProton:
Fixes:
+ CRASH - Importing from the Library during a simulation causes a crash
+ CRASH - Adding a Cylinder Crashes the Program after running Molecular Emitter

### DIFF
--- a/godot_project/editor/controls/dockers/workspace_docker/groups_docker/groups_docker_controls/nano_groups_list.gd
+++ b/godot_project/editor/controls/dockers/workspace_docker/groups_docker/groups_docker_controls/nano_groups_list.gd
@@ -385,6 +385,8 @@ func _on_workspace_context_current_structure_context_changed(in_structure_contex
 
 func _on_workspace_context_selection_in_structures_changed(in_structure_contexts: Array[StructureContext]) -> void:
 	for context: StructureContext in in_structure_contexts:
+		if not context.nano_structure:
+			continue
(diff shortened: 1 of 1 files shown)


godot_project/editor/controls/dockers/workspace_docker/groups_docker/groups_docker_controls/nano_groups_list.gd
```diff
@@ -385,6 +385,8 @@ func _on_workspace_context_current_structure_context_changed(in_structure_contex
 
 func _on_workspace_context_selection_in_structures_changed(in_structure_contexts: Array[StructureContext]) -> void:
 	for context: StructureContext in in_structure_contexts:
+		if not context.nano_structure:
+			continue
```
MarianoGnu, line 389: this looks fishy, why is it happening in first place? looks to me this context shouldn't be in the list, or at least there should be a comment explaining why this happened.

There could be more callbacks connected to this signal that could have this same issue

HungryProton: I don't have a full stack trace as this is one of the flushed signals. My guess is, this happens during a snapshot and the state is inconsistent, but I haven't found the root cause yet

MarianoGnu:
I created https://github.com/MSEP-one/msep.one/pull/289 as counter proposal of this

PR #291, opened 2025-07-01
Ms/emitter crash

MarianoGnu:
Task: BUG: Assertion failed: Attempting to create instances when instances already exists!


MarianoGnu:
that Revert was not meant to be part of the PR

PR #299, opened 2025-07-03
Fix crash when initializing the WorkspaceCamera

MarianoGnu:
This was an oversight on https://github.com/MSEP-one/msep.one/pull/297

Task: Lines of Virtual Objects not visible from inside the object


MarianoGnu:
Proactively merged since this crash was too intrusive

PR #319, opened 2025-07-22
Dont abort creating reference shapes when leaving the shapes RingMenuLevel

MarianoGnu:
Task: BUG: Is possible to break the **Create reference shape** functionality


HungryProton (approved):
I'm a bit worried this used to be here because of another bug, but I can't find the original commit and it doesn't seem to cause any issues

MarianoGnu:
> I'm a bit worried this used to be here because of another bug, but I can't find the original commit and it doesn't seem to cause any issues

I remember writing this code, it simnply made sense in my mind at the time if entering the menu starts creating the object, then exiting the menu aborts it
There was not any other strong reason for aborting

PR #321, opened 2025-07-23
Fix structure preview not showing after changing the active group

HungryProton:
Fixes: BUG: Small Molecules preview dissapears


MarianoGnu:
I found a corner case not covered by this PR:


https://github.com/user-attachments/assets/2fdb8477-92a6-48d0-869e-566c1ed37f10

PR #333, opened 2025-07-28
Fix the camera dialogue

HungryProton:
Fixes: BUG: Camera altitude control rolls instead of yaws

### DIFF
--- a/godot_project/editor/controls/dockers/workspace_docker/c_dynamic_context_docker/dynamic_context_controls/inspector_controls/inspector_control_direction/inspector_control_direction.gd
+++ b/godot_project/editor/controls/dockers/workspace_docker/c_dynamic_context_docker/dynamic_context_controls/inspector_controls/inspector_control_direction/inspector_control_direction.gd
@@ -13,7 +13,7 @@ enum Mode {
 var _preview: Control
 var _altitude_spin_box: SpinBoxSlider
 var _azimuth_spin_box: SpinBoxSlider
-
+var _tilt_spin_box: SpinBoxSlider
 
 var _property_changed_signal := Signal()
 var _getter := Callable()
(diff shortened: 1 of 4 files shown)


godot_project/editor/controls/dockers/workspace_docker/c_dynamic_context_docker/dynamic_context_controls/inspector_controls/inspector_control_direction/inspector_control_direction.gd
```diff
@@ -87,7 +87,7 @@ func _preview_draw() -> void:
 	_preview.draw_set_transform(drawn_rect.get_center(), 0.0, Vector2(azimuth_dir.x, 1.0))
 	_preview.draw_arc(ORIGIN, smaller_size * 0.5, -PI * 0.5, PI * 0.5, int(POINT_COUNT * 0.5), Color.WHITE)
 	# line indicating altitude
-	var altitude_point: Vector2 = (azimuth_point * Vector2(1.0, 0.3)).normalized()
+	var altitude_point: Vector2 = (Vector2.RIGHT * Vector2(1.0, 0.3)).normalized()
```
HungryProton: Not directly related, but this fix the visual on the right. It shows the wrong altitude direction, but only when the azimuth is negative

MarianoGnu (changes requested):
Still not quite right, after applyying the change, closing the dialog with a non 0 azimut, and reopening it, the value of azimuth changes to a wrong value


https://github.com/user-attachments/assets/ede44de3-db82-4197-a7cf-4c6c0758d768

HungryProton:
Updated the PR to also add tilt support, otherwise it causes a camera jump when changing the values (since the tilt is automatically reset to 0 if we only have azimuth and altitude)

MarianoGnu (commented):
sheep it

MarianoGnu (approved):

PR #352, opened 2025-08-07
Ensure the system knows particles exists at all times

MarianoGnu:
- When performing a 'whitelisted' action, the system injects the action to undo stack, but like if there where no simulation
- Then returns to the simulated state
- While performing this operation, the system 'forgot' some particles created by particle emitter exists

Task: BUG: Only the first molecule created by particle emitter is selectable

### DIFF
--- a/godot_project/project_workspace/workspace_context/history/history.gd
+++ b/godot_project/project_workspace/workspace_context/history/history.gd
@@ -90,14 +90,18 @@ func register_snapshotable(in_system: Object) -> void:
 	_snapshotable_systems.append(in_system)
 
 
-func create_snapshot(in_snapshot_name: String) -> Dictionary:
-	_monitor_redundant_snapshots(in_snapshot_name)
+func create_snapshot_no_push() -> Dictionary:
 	var snapshot: Dictionary = {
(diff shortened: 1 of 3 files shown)


godot_project/project_workspace/workspace_context/history/history.gd
```diff
@@ -90,14 +90,18 @@ func register_snapshotable(in_system: Object) -> void:
 	_snapshotable_systems.append(in_system)
 
 
-func create_snapshot(in_snapshot_name: String) -> Dictionary:
-	_monitor_redundant_snapshots(in_snapshot_name)
+func create_snapshot_not_push() -> Dictionary:
```
HungryProton: nit: Name should be `create_snapshot_no_push` (without the t)
or maybe it could just be an optional argument `add_to_stack` that default to true in the `create_snapshot` function

HungryProton (approved):

godot_project/project_workspace/workspace_context/history/history.gd
MarianoGnu: i preffer to have 2 methods for being more explicit, in line with the "Single Purpose" principle

PR #359, opened 2025-08-08
Hide the ghost shape preview when picking a new one from the docker

HungryProton:
Fixes: BUG - Shapes do strange animation when switching shape types


MarianoGnu (approved):
It works, sheep it 🐑

PR #386, opened 2025-08-20
Show warnings when particle emitters are overlapping volume in it's initial position

MarianoGnu:
- Also show an special message if one of the emitters is hidden, so user can know of this situation

Task: BUG - Leads to Crash - Emitters

image

image

### DIFF
--- a/godot_project/utils/workspace_utils.gd
+++ b/godot_project/utils/workspace_utils.gd
@@ -153,6 +153,10 @@ static func has_emitters_with_potentially_unstable_parameters(in_workspace_conte
 	return _has_emitters_with_potentially_unstable_parameters(in_workspace_context)
 
 
+static func has_overlapping_emitters(in_workspace_context: WorkspaceContext, in_search_for_hidden: bool) -> bool:
+	return _has_overlapping_emitters(in_workspace_context, in_search_for_hidden)
+
+
 static func structure_context_has_drastically_invalid_tetrahedral_structure(in_structure_context: StructureContext, in_atoms: PackedInt32Array) -> bool:
 	if not in_structure_context.nano_structure is AtomicStructure:
 		return false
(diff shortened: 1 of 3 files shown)


godot_project/utils/workspace_utils.gd
```diff
@@ -205,6 +209,25 @@ static func _has_emitters_with_potentially_unstable_parameters(in_workspace_cont
...
+		for j in range(i + 1, emitters.size()):
+			var j_emitter: NanoParticleEmitter = emitters[j]
+			var j_aabb: AABB = j_emitter.get_aabb(AtomicStructure.AABB_BoundsType.ContactRadius)
+			if i_aabb.intersects(j_aabb):
+				if in_search_for_hidden:
+					var j_hidden: bool = not j_emitter.get_visible()
+					if i_hidden or j_hidden:
+						return true
+				else:
+					return true
```
HungryProton, line 227: The logic here could be simplified and sped up.

On line 217, if the first emitter is hidden and `in_search_for_hidden` is false, you can skip the second loop completely.

Same in the J loop, check for visibility first and if `in_search_for_hidden` is false, you can skip the intersection check early.

HungryProton (approved):

godot_project/utils/workspace_utils.gd
MarianoGnu: > On line 217, if the first emitter is hidden and in_search_for_hidden is false, you can skip the second loop completely.

The point of this function is searching for overlapping emitters, not searching for hidden emitters, so no, i cannot skip the second loop

HungryProton: It search for overlapping emitters, but only among the visible ones, unless `in_search_for_hidden` is true, right?
The optimization I'm talking about only applies when that parameter is false

MarianoGnu: i see!

MarianoGnu: I give this a little more tough, turns out i cannot skip the loop when emitters are not hidden, because i also care for a case when `i` is visible, but `j` is not
If i skip `i` then that means `j` should iterate from `0 to size` instead of `i+1 to size` to account for that situation, makes not worth the optimization

PR #397, opened 2025-08-26
Update how particle emitters work internally

HungryProton:
Fixes: CRASH: deleting particle emitter during simulation
and potentially every other emitter bugs related to undo redo.

---

### Current state
+ Emitters create all their instances when a simulation is started, and deletes them when the simulation ends.
+ When an action (like changing selection) happens during a simulation, the target atom might not exists once the simulation ends.
  - Redoing these actions leads to various bugs.
  - A lot of code exists to handle these edge cases, but doesn't cover all scenarios 

### This PR
+ Emitters create their instances as soon as possible
  - From the create emitter panel.
  - When the file is loaded from disk.
+ Actions happening during a simulation will target atoms that already existed before the simulation (or after it ended).
+ Redoing these actions will work
  - Some atoms might be marked as invalid, in that case, nothing happens but it's still marked as an undo / redo step.
+ No longer need special code paths for emitters (except in openMM)


godot_project/autoloads/feature_flag_manager/FeatureFlagManager.gd
```diff
@@ -25,8 +25,8 @@ const RELAX_EDITABLE_TEMPERATURE = &"feature_flags/relax_editable_temperature"
 const TEMPERATURE_IN_FAHRENHEIT = &"feature_flags/temperature_in_fahrenheit"
 const FEATURE_FLAG_VIRTUAL_MOTORS = &"feature_flags/virtual_motors"
 const FEATURE_FLAG_PARTICLE_EMITTERS = &"feature_flags/particle_emitters"
-const FEATURE_FLAG_EMITTERS_WITH_EMMIT_COUNT = &"feature_flags/particle_emitters_with_emmit_count"
-const FEATURE_FLAG_EMITTERS_SHOW_UNSPAWNED_INSTANES = &"feature_flags/particle_emitters_show_unspawned_instanes"
```
HungryProton, line 29: This fixes a typo (emmit instead of emit), not directly related to the fix

godot_project/editor/controls/dockers/workspace_docker/a_create_docker/create_particle_emitter_panel.gd
```diff
@@ -199,6 +199,7 @@ func _create_particle_emitter_from_template(out_template: AtomicStructure, in_em
 		])
 		out_emitter.set_parameters(instance_parameters)
 		out_emitter.set_position(in_emitter_pos)
+		out_emitter.ensure_instances_exists()
```
HungryProton: creates all instances when adding the emitter to the workspace

godot_project/project_workspace/workspace_context/structure_context/collision_engine/collision_space/collision_space.gd
```diff
@@ -73,6 +73,8 @@ func add_collider(in_object_id: int, in_shape_id: RID, in_transform: Transform3D
 
 
 func free_collider_and_get_related_shape(in_object_id: int) -> RID:
+	if not _id_to_collision_rid.has(in_object_id):
+		return RID()
```
HungryProton: **Feedback needed**

`ensure_instances_exists` creates and deletes atoms during the same frame, so the renderer and collision engine try to act on something they're not tracking yet.

Either we accept to make these 2 APIs a bit less robust, or we need to add a way to create invalid atoms in a single action in the `NanoMolecularStructure`

godot_project/project_workspace/workspace_context/workspace_context.gd
```diff
@@ -626,15 +624,14 @@ func end_simulation_if_running() -> void:
 
 ## This method is automatically called when any code attempts
 ## to create a new action with `snapshot_moment()`
-func apply_simulation_if_running() -> void:
+func _apply_simulation_if_running() -> void:
```
HungryProton: Calling this directly will bypass the `about_to_apply_simulation` signal and break the emitters. So this is private now.

godot_project/utils/segmented_multimesh/segmented_multi_mesh.gd
```diff
@@ -311,6 +311,8 @@ func update_particle_color(in_external_id: int, new_color: Color, in_additional_
 
 ## Need to call apply_queued_removals() after all changes are queued
 func queue_particle_removal(in_external_id: int) -> void:
+	if not _external_id_to_particle.has(in_external_id):
+		return
```
HungryProton: Same reason as in `collision_space.gd` above

godot_project/project_workspace/structs/nano_particle_emitter.gd
```diff
@@ -123,33 +180,47 @@ func create_instances(out_group: AtomicStructure) -> void:
...
+	_instances_group.start_edit()
+	for instance: PackedInt32Array in _instances_atom_ids:
+		for atom_id: int in instance:
+			if _instances_group.is_atom_valid(atom_id):
+				_instances_group.remove_atom(atom_id)
+	for instance: PackedInt32Array in _instances_bond_ids:
+		for bond_id: int in instance:
+			if _instances_group.is_bond_valid(bond_id):
+				_instances_group.remove_bond(bond_id)
+	_instances_group.end_edit()
```
HungryProton: Making all atoms invalid during the same step as we're creating them doesn't work.
It works when flushing the previous commands and starting to edit again, but maybe it's worth creating a dedicated api for that?

godot_project/project_workspace/workspace_context/workspace_context.gd
MarianoGnu: nit: probably should convert ## comments to # since is no longer a public API

godot_project/project_workspace/structs/nano_particle_emitter.gd
MarianoGnu: remove_atom/bond could check if the atom_id is in the "added" queue and remove it from there instead of adding it to the removed queue

godot_project/autoloads/MolecularEditorContext.gd
```diff
@@ -170,7 +170,7 @@ func load_and_activate_workspace(in_path: String) -> void:
 		return
 	var workspace_context: WorkspaceContext = get_workspace_context(workspace)
 	workspace_context.set_camera_global_transform(workspace.camera_transform)
-	# Needs to call_deferred to be executed after msep_editor_settings.changed is emmited
+	# Needs to call_deferred to be executed after msep_editor_settings.changed is emited
```
MarianoGnu: typo emitted

godot_project/autoloads/openmm/openmm.gd
```diff
@@ -348,8 +348,7 @@ func _request_start_simulation(
 	const NUDGE_ATOMS_FIX = false
 	# Before creating the payload, prewarm the atoms of particle emitters
 	for emitter: NanoParticleEmitter in in_workspace_context.get_particle_emitters():
-		var parent: NanoStructure = in_workspace_context.workspace.get_parent_structure(emitter)
-		emitter.create_instances(parent)
+		emitter.revalidate_all_instances()
```
MarianoGnu, line 351: i am concerned about the fact this could be needed here, if instances are created in this moment is already too late, make make an assert instead

godot_project/editor/controls/dockers/workspace_docker/simulations_docker/simulations_docker.gd
```diff
@@ -84,7 +84,8 @@ func _on_create_object_parameters_simulation_type_changed(_in_new_type: CreateOb
 	if context == null:
 		return
 	# This will flush the state of any running simulation and terminate it
-	context.apply_simulation_if_running()
+	if context.is_simulating():
+		context.snapshot_moment(tr("Apply Simulation State"))
```
MarianoGnu: nit: this should probably pause the simulation and show a prompt for confirmation, but that should be a sepparate PR after confirming with jonathan

godot_project/project_workspace/workspace_context/structure_context/collision_engine/collision_space/collision_space.gd
MarianoGnu: same answer, we can possibly make the API remove the atom/bonds IDs from the added queue instead of adding the id to both added and removed queue

godot_project/project_workspace/workspace_context/workspace_context.gd
```diff
@@ -1412,38 +1409,19 @@ func snapshot_moment(in_operation_name: String) -> void:
...
-			await get_tree().process_frame
-		
 		# Rewind simulation to the latest point and create a snapshot.
 		# This will drop the user operation (final_snapshot) from history. 
 		seek_simulation(current_simulation_time)
-		apply_simulation_if_running()
-		end_simulation_if_running()
-		
+		_apply_simulation_if_running()
+		_history.create_snapshot(tr("Apply Simulation State"))
```
MarianoGnu: :warning: I dont see any check confirming if `in_operation_name == "Apply Simulation State"` so this is potentially creating 2 snapshots?

godot_project/project_workspace/structs/nano_particle_emitter.gd
```diff
@@ -371,16 +415,63 @@ func apply_state_snapshot(in_state_snapshot: Dictionary, in_with_instances: bool
...
+
+
+func _on_workspace_loaded(in_workspace: Workspace) -> void:
+	if _instances_group:
+		# Already initialized
+		MolecularEditorContext.workspace_loaded.disconnect(_on_workspace_loaded)
+		return
+	if in_workspace.has_structure(self):
+		initialize(in_workspace)
+		MolecularEditorContext.workspace_loaded.disconnect(_on_workspace_loaded)
```
MarianoGnu: i believe there are better places to do this initialization, in example from `WorkspaceContext._on_workspace_main_view_ready` 

initialize could be private, and just be called from `ensure_instances_exists` and not return on line 104

godot_project/project_workspace/structs/nano_particle_emitter.gd
```diff
@@ -12,29 +18,46 @@ const DEFAULT_TRANSFORM = Transform3D(Basis(DEFAULT_ROTATION))
...
 	var on_feature_flag_toggled: Callable = func(path: String, new_value: bool) -> void:
-		if path == FeatureFlagManager.FEATURE_FLAG_EMITTERS_SHOW_UNSPAWNED_INSTANES:
+		if path == FeatureFlagManager.FEATURE_FLAG_EMITTERS_SHOW_UNSPAWNED_INSTANCES:
 			_debug_show_unspawned_instances = new_value
 	FeatureFlagManager.on_feature_flag_toggled.connect(on_feature_flag_toggled)
+	MolecularEditorContext.workspace_loaded.connect(_on_workspace_loaded)
+
 
+func initialize(in_workspace: Workspace) -> void:
+	if not in_workspace:
```
MarianoGnu: can this happen? please add a comment explaining in wich situation initialize could be called with `null`

godot_project/project_workspace/structs/nano_particle_emitter.gd
```diff
@@ -69,18 +92,54 @@ static func calculate_total_molecule_instance_count(
...
+	# If the emitter moved since previous instances were created, reposition the existing ones
+	# before creating the missing instances.
+	var reference_inverse: Transform3D = _instances_reference_transform.affine_inverse()
+	if not _transform.is_equal_approx(_instances_reference_transform):
+		_instances_group.start_edit()
+		for instance: PackedInt32Array in _instances_atom_ids:
+			for atom_id: int in instance:
+				var local_pos: Vector3 = reference_inverse * _instances_group.atom_get_position(atom_id)
+				_instances_group.atom_set_position(atom_id, _transform * local_pos)
+		_instances_group.end_edit()
```
MarianoGnu: isn't this transformation something that should always happen on `set_transform` and not only when generating instances

godot_project/autoloads/openmm/openmm.gd
MarianoGnu: hum i see this method does a little more than ensuring instances exists, but i still think it should asset instances exists inside of it instead of creating them if needed

godot_project/project_workspace/structs/nano_particle_emitter.gd
```diff
@@ -371,16 +415,63 @@ func apply_state_snapshot(in_state_snapshot: Dictionary, in_with_instances: bool
...
+		MolecularEditorContext.workspace_loaded.disconnect(_on_workspace_loaded)
+		return
+	if in_workspace.has_structure(self):
+		initialize(in_workspace)
+		MolecularEditorContext.workspace_loaded.disconnect(_on_workspace_loaded)
+
+
+
+func _on_parameters_changed() -> void:
+	ensure_instances_exists()
```
MarianoGnu, line 475: ⚠️  this is connected to ensure there's enough instances if instance rate or some other parameter changes, but it is also needed to confirm instances exists if the lenght of the simulation changes, because if there's not fixed limit for emission more instances could be needed

godot_project/project_workspace/structs/nano_particle_emitter.gd
```diff
@@ -371,16 +415,63 @@ func apply_state_snapshot(in_state_snapshot: Dictionary, in_with_instances: bool
...
 		var workspace_context: WorkspaceContext = MolecularEditorContext.get_workspace_context(workspace)
-		var group_structure_context: StructureContext = \
-			workspace_context.get_nano_structure_context_from_id(group_id)
-		_instances_group = group_structure_context.nano_structure as AtomicStructure
-		_instances_group.apply_state_snapshot(in_state_snapshot["_instances_group_state"])
-		group_structure_context.get_collision_engine().rebuild(group_structure_context)
-		var rendering: Rendering = workspace_context.get_rendering()
-		var renderer := rendering._get_renderer_for_atomic_structure(_instances_group)
-		renderer.apply_state_snapshot(in_state_snapshot["_instances_group_renderer_state"])
+		workspace_context.about_to_apply_simulation.disconnect(_on_about_to_apply_simulation)
```
MarianoGnu, line 512: hummm this signal is removed on deletion, but if you undo and redo deletion i believe nothing is initializing any of the signals connected on `initialize`, at least until something calls `ensure_instances_exists`

MarianoGnu (changes requested):
Changes looks good to me, but i have a few questions before approval

godot_project/autoloads/openmm/openmm.gd
HungryProton: This is mostly to make them active again (and at the right place) rather than creating the instances. It calls `ensure_instances_exists` because the repositioning logic is there but I can change it if needed. Or I could auto reposition the instances every time the emitter is moved. It's a bit wasteful but at least it'll be more robust.

godot_project/project_workspace/workspace_context/workspace_context.gd
HungryProton: Oh that's interesting, we already create 2 snapshots on purpose (one to apply the simulation, the other to apply the user operation), so that's normal and expected.

But if the user is only applying the simulation, then there's no need to do all of that and it can be skipped. I'll fix it.

MarianoGnu: there's an "Apply Simulation button"

If i am correct, cliking it will run `snapshot_moment("Apply Simulation")` that will also trigger `_history.create_snapshot("Apply simulation")` here and `_history.push_and_apply_snapshot(final_snapshot, in_operation_name)` in line 1426 with `in_operation_name == "Apply Simulation State"`

And this will produce no error or warning because the system to detect redundant snapshots ignores `"Apply Simulation State"`

It would just add 2 identical snapshots to undo history

HungryProton: Yes that's the edge case I'm handling (I haven't pushed the changes yet)

HungryProton: Fixed above on lines 1405-1412

godot_project/project_workspace/workspace_context/workspace_context.gd
HungryProton: Done

godot_project/project_workspace/structs/nano_particle_emitter.gd
HungryProton: Yes that works!

godot_project/autoloads/MolecularEditorContext.gd
HungryProton: Fixed

godot_project/autoloads/openmm/openmm.gd
HungryProton: Changed it to only revalidate the atoms. There's also an assert to ensure all instances already exists.

godot_project/project_workspace/structs/nano_particle_emitter.gd
HungryProton: I've added a `notify_added_to_workspace` method to `NanoStructure`, so the initialization now happens exactly once when everything is ready.

godot_project/project_workspace/structs/nano_particle_emitter.gd
HungryProton: Can't happen anymore.

godot_project/project_workspace/structs/nano_particle_emitter.gd
HungryProton: Moved the logic to `set_transform`

godot_project/project_workspace/structs/nano_particle_emitter.gd
HungryProton: Added a connection to `simulation_parameters.changed`

godot_project/project_workspace/structs/nano_particle_emitter.gd
HungryProton: This was replaced with a bool flag instead (that reset on apply_snapshot). So the connection stays there, but is ignored only if the emitter is about to be deleted during a simulation.

godot_project/project_workspace/workspace_context/workspace_context.gd
HungryProton: I reverted it to be public in the end, it made more sense to emit the `about_to_apply` signal from here instead.

godot_project/editor/controls/dockers/workspace_docker/simulations_docker/simulations_docker.gd
HungryProton: This was reverted

godot_project/project_workspace/structs/nano_particle_emitter.gd
HungryProton: Signals are disconnected on remove now, and recreated each time the structure is added back to the workspace

HungryProton:
All the changes are addressed I think, and I could not reproduce any of the issues that appeared during development.

godot_project/project_workspace/workspace.gd
```diff
@@ -162,7 +162,7 @@ var _id: int
 
 
 func _init() -> void:
-	# prevent changed and representation_settings_changed signals from being emmited
+	# prevent changed and representation_settings_changed signals from being emited
```
MarianoGnu: typo emitted

godot_project/project_workspace/workspace_context/workspace_context.gd
```diff
@@ -1404,57 +1405,43 @@ func snapshot_moment(in_operation_name: String) -> void:
...
 		
 		# apply overdue signals, ensure snapshots contain up to date state
 		_emit_selection_in_structures_changed()
 		_emit_structure_content_changed()
 		
+		# Special case if the user only applied the simulation, a single snapshot is enough.
+		if in_operation_name == tr("Apply Simulation State"):
+			apply_simulation_if_running()
+			_is_applying_simulation_state = false
+			simulation_state_applied.emit()
```
MarianoGnu, line 1417: this signal is only emitted when "Apply Simulation State" is executed by changing something during simulation.
I checked the usage and seems to only matter in this situation, and in fact emitting it inside `apply_simulation_if_running()` before setting `_is_applying_simulation_state = false` would be wrong.
I have no strong opinion about what to do about it, i just hope nobody connects to it without knowing how it works

MarianoGnu (approved):
All known crashes are gone \o/
I have a last observation, but is a non blocker

PS: This PR also fixes the repro steps found for **CRASH - Doing Undo While testing Hover Group Mismatch bug**

godot_project/project_workspace/workspace_context/workspace_context.gd
HungryProton: I'm not really sure what to do about it either, but since the only script using it is nano_group_list, we can probably address that in another PR if necessary

godot_project/project_workspace/workspace.gd
HungryProton: fixed

PR #399, opened 2025-08-27
Fix an issue when the workspace contains a single atom centered on the origin

HungryProton:
Fixes: CRASH: placing a single atom at the center of the workspace

The old code assumes the default AABB() means the workspace is empty, which is not true if there's a single atom at (0,0,0)

### DIFF
--- a/godot_project/project_workspace/workspace_context/structure_context/structure_context.gd
+++ b/godot_project/project_workspace/workspace_context/structure_context/structure_context.gd
@@ -257,6 +257,12 @@ func has_atom_selection(include_children: bool = true) -> bool:
 	return false
 
 
+func is_empty() -> bool:
+	if nano_structure is AtomicStructure:
+		return nano_structure.get_valid_atoms_count() == 0
+	return false
+
+
 func is_empty_but_has_subgroups() -> bool:
 	return nano_structure is AtomicStructure and \
 			nano_structure.get_valid_atoms_count() == 0 and \
--- a/godot_project/utils/workspace_utils.gd
+++ b/godot_project/utils/workspace_utils.gd
@@ -703,11 +703,15 @@ static func _get_visible_objects_aabb(out_workspace_context: WorkspaceContext) -
 	var visible_objects_aabbs: Array[AABB] = []
 	var visible_structure_contexts: Array[StructureContext] = out_workspace_context.get_visible_structure_contexts()
 	for context in visible_structure_contexts:
-		var aabb: AABB = context.nano_structure.get_aabb().abs()
-		if aabb == AABB():
-			# structure is empty
+		if context.is_empty():
 			continue
-		visible_objects_aabbs.push_back(context.nano_structure.get_aabb().abs())
+		var aabb: AABB
+		if context.nano_structure is AtomicStructure:
+			var atomic_structure: AtomicStructure = context.nano_structure
+			aabb = atomic_structure.get_aabb(AtomicStructure.AABB_BoundsType.VisualRadius).abs()
+		else:
+			aabb = context.nano_structure.get_aabb().abs()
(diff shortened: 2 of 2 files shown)


godot_project/utils/workspace_utils.gd
```diff
@@ -703,10 +703,9 @@ static func _get_visible_objects_aabb(out_workspace_context: WorkspaceContext) -
 	var visible_objects_aabbs: Array[AABB] = []
 	var visible_structure_contexts: Array[StructureContext] = out_workspace_context.get_visible_structure_contexts()
 	for context in visible_structure_contexts:
-		var aabb: AABB = context.nano_structure.get_aabb().abs()
-		if aabb == AABB():
-			# structure is empty
+		if context.is_empty():
 			continue
+		var aabb: AABB = context.nano_structure.get_aabb().abs()
```
MarianoGnu: Note, override for `AtomicStructure.get_aabb` has an override with an additional argumentthat allows to calculate the size including visual or contact radius of atoms:
```
enum AABB_BoundsType {
	AtomsPositions, # get_aabb will not grow based on the size of atoms
	VisualRadius,   # get_aabb will grow based on how big atoms are rendered
	CovalentRadius, # get_aabb will grow based on each atom's covalent radius
	ContactRadius,  # get_aabb will grow based on each atom's contact radius
}
```
This could workaround the issue as well, and given the context of this method probably `AABB_BoundsType.VisualRadius` makes sense

MarianoGnu (approved):

godot_project/project_workspace/workspace_context/structure_context/structure_context.gd
```diff
@@ -257,6 +257,12 @@ func has_atom_selection(include_children: bool = true) -> bool:
 	return false
 
 
+func is_empty() -> bool:
+	if nano_structure is NanoMolecularStructure:
```
MarianoGnu: ⚠️  this should be `is AtomicStructure` otherwise will fail for `LMDBNanoStruct`

MarianoGnu (changes requested):
Actually, found a real issue

godot_project/project_workspace/workspace_context/structure_context/structure_context.gd
HungryProton: Fixed

godot_project/utils/workspace_utils.gd
HungryProton: I didn't know this one, I've added it to the fix

godot_project/utils/workspace_utils.gd
```diff
@@ -703,10 +703,14 @@ static func _get_visible_objects_aabb(out_workspace_context: WorkspaceContext) -
...
-			# structure is empty
+		if context.is_empty():
 			continue
+		var aabb: AABB
+		if context.nano_structure is AtomicStructure:
+			var atomic_structure: AtomicStructure = context.nano_structure
+			aabb = atomic_structure.get_aabb(AtomicStructure.AABB_BoundsType.VisualRadius).abs()
+		else:
+			aabb = context.nano_structure.get_aabb().abs()
 		visible_objects_aabbs.push_back(context.nano_structure.get_aabb().abs())
```
MarianoGnu: you are creating an `aabb` variable, but you are still pushing to the array `context.nano_structure.get_aabb().abs()`

MarianoGnu (changes requested):
one more change

godot_project/utils/workspace_utils.gd
HungryProton: Oh I didn't even notice since it was already the case before, fixed!

MarianoGnu (approved):

PR #434, opened 2025-09-18
Don't create an undo step when grow / shrink selection can't change the selection

HungryProton:
Fix: BUG: "Grow Selection" and "Select Connected" adds an undo step even when nothing new was selected, also is available from visual/topbar menu but not from Ring Menu


MarianoGnu (approved):
Task referred to "Grow Selection" and "Select Connected", not shrink selection

This PR looks good, but only solves half of the issue

HungryProton:
Added the fix for selected connected too.

MarianoGnu (approved):

PR #438, opened 2025-09-22
Fix highlighting of 'Find Visible Atoms by Type' from TopBar/VisualMenu

MarianoGnu:
- Panel was renamed, but the code of the menu was outdated
- Also added an assert to more easily discover this problem in the future

Task: BUG: Using "Select atoms by type" from visual/topbar menu does not highlight the control in the dynamic context panel


HungryProton (approved):

HungryProton:
That works but it won't prevent the issue from happening again.
We should probably define these magic strings in a single place and reference them from that source only in a future PR

MarianoGnu:
> That works but it won't prevent the issue from happening again. We should probably define these magic strings in a single place and reference them from that source only in a future PR

This will now assert if you try to use an undefined string. We can create and use constants, but we cannot warrantee such constants will be used in the future, in the end human error is unavoidable.

PR #455, opened 2025-10-08
Fix ffmpeg generating unplayable mp4 videos when a non standard resolution is used

MarianoGnu:
- This change makes ffmpeg to always generate frames with even sizes

Before this, if you used the crop tool, and selected an area with an odd size (in example 918x587) that would make ffmpeg to generate a video that is unplayable because of some known limitation on H.264 and rbg data


HungryProton (changes requested):
ffmpeg crashes on my end and complains this option does not exists.

Since you're resizing the image on line 119 already, why not force _resolution to have an even number directly?

HungryProton (approved):

PR #463, opened 2025-10-20
Add a _file_format_version field to Workspace files

MarianoGnu:
- This could allow us to react when loading a version of a file that doesn't match expectations
- An older file may be not supported, or need an upgrade on load
- Files newer to the current version of MSEP could show a warning to the user requiring to download a newer version of MSEP

Task; Add version flag to the msep1 files

### DIFF
--- a/godot_project/project_workspace/file_format/workspace_format_loader.gd
+++ b/godot_project/project_workspace/file_format/workspace_format_loader.gd
@@ -42,4 +42,11 @@ func _load(path: String, _original_path: String, _use_sub_threads: bool, cache_m
 		if workspace != null:
 			workspace.take_over_path(path)
 	d.remove(tmp_path)
+	if workspace._file_format_version < Workspace.get_most_recent_file_format_version():
+		push_warning("This workspace file was written with an older file format version: ",
+			Workspace.FileFormatVersion.find_key(workspace._file_format_version))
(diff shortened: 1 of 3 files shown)


godot_project/project_workspace/file_format/workspace_format_loader.gd
```diff
@@ -42,4 +42,11 @@ func _load(path: String, _original_path: String, _use_sub_threads: bool, cache_m
 		if workspace != null:
 			workspace.take_over_path(path)
 	d.remove(tmp_path)
+	if workspace._file_format_version < Workspace.get_most_recent_file_format_version():
```
HungryProton, line 45: If I load an older file, this script fails even earlier, on line 41, since workspace is null.
The version number should probably be extracted manually at the start of this function, instead of relying on ResourceLoader to work.

HungryProton: The whole save format should not be a raw Godot resource to begin with but I don't what are the plans for that

MarianoGnu: - workspace will only be null if the file is so old it fails to load the structure list
- file format will inevitably  change when we implement LMDB file structure

HungryProton (approved):

PR #468, opened 2025-10-22
Use full path to ffmpeg as part of the recording command

MarianoGnu:
- This fixes a crash on macos because execute_with_pipe cannot run the command because the lack of the $PATH environment variable

Task: CRASH - Saving MP4 files on mac


MarianoGnu:
>Note: On macOS, sandboxed applications are limited to run only embedded helper executables, specified during export or system .app bundle, system .app bundles will ignore arguments.

We cannot trust the system to resolve the file path to the executable, so we will change the approach to embedding ffmpeg

PR #474, opened 2025-10-29
Keep the outdated warning consistent with the undo redo history

HungryProton:
Fixes: BUG - Project announces it has been changed since last validation, but validation has never been run

This PR does 2 things:

+ It makes the validation panel a snapshotable system (so it can track when to show the warning label)
+ It changes how history version is calculated

Currently, version is a simple int, it goes up when an action happens, and down on undo, which leads to version numbers being reused for different snapshots.

```
Snap A -> v1
Snap B -> v2
Undo (Back to A) -> v1
Snap C -> v2  (but it should be v3)
```

Instead, this PR adds a new array to track the snapshot versions (just like the names), it's increased on each new snapshot, reused on redo, but new actions created after an undo will resume from the highest version number.

### DIFF
--- a/godot_project/project_workspace/workspace_context/history/history.gd
+++ b/godot_project/project_workspace/workspace_context/history/history.gd
@@ -80,8 +80,9 @@ var _workspace_context: WorkspaceContext
 var _snapshotable_systems: Array[Object]
 var _snapshot_stack: Array[Dictionary]
 var _name_stack: Array[String] = []
+var _version_stack: Array[int] = []
 var _stack_pointer: int = -1
-var _version: int
+var _highest_version: int = -1
 var _last_snapshot_taken_at_frame: int = Engine.get_frames_drawn() - 1
 var _last_snapshot_name: String = ""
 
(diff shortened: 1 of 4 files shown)


godot_project/project_workspace/workspace_context/history/history.gd
```diff
@@ -80,8 +80,9 @@ var _workspace_context: WorkspaceContext
 var _snapshotable_systems: Array[Object]
 var _snapshot_stack: Array[Dictionary]
 var _name_stack: Array[String] = []
+var _version_stack: Array[int] = []
 var _stack_pointer: int = -1
-var _version: int
+var _highest_version: int = -1
```
HungryProton, line 85: I don't like adding yet another array to track a parameter that should be attached to the snapshot directly but I'm not sure if we can add random data to a snapshot without affecting the rest.

MarianoGnu (approved):
I am under the impression the version stack wasn't fully necesary, however the code is good to go

HungryProton:
> I am under the impression the version stack wasn't fully necesary, however the code is good to go

I was hoping to fix the issue without adding a snapshot system at all and just checking the version, but it caused other issues and both had to be done in the end.

PR #480, opened 2025-10-30
Update submodules to include the pipe fix & new export templates

HungryProton:
Fixes: CRASH - Attempting to save a video twice crashes the app

Windows files are unaffected as only os_unix.cpp was modified.


MarianoGnu:
Just to confirm, macos is built as a fat binary for both x86 and arm64 archs?

MarianoGnu (approved):

HungryProton:
Yes, it contains both

PR #482, opened 2025-10-31
fix incorrect selection state after opening the 'Capture camera image' popup

HungryProton:
Fixes: BUG: Opening the "Capture Camera Image" makes several sections of the application believe there's no selection


MarianoGnu (approved):
Smart!

PR #486, opened 2025-11-04
Make the screen capture respect the theme and color overrides

HungryProton:
Card: BUG: in "Capture Camera Image" dialog, the "Environment" background overrides the Academic View background


MarianoGnu (approved):
sheep it 🐑

PR #497, opened 2025-11-11
Limit the position of atoms to remain in the -3000/+3000 nanometers of the origin

MarianoGnu:
Task: Limit camera and objects positions


MarianoGnu:
Note to self: This task turned to be absurd, even limiting the positions of atoms and camera, is still basically impossible to fit the existing atoms in the oposite corner of the world, because of the FOV being so small. Maybe is simply not worth merging and accept this is something that can happen

PR #499, opened 2025-11-12
Solved a corner case in 'Fix overlapping atoms'

MarianoGnu:
- All atoms was getting the same 'direction from center' because of rounding presition issue

Task: BUG - Correct Overlapping Atoms Doesn't work

### DIFF
--- a/godot_project/editor/controls/dockers/workspace_docker/simulations_docker/simulations_docker_controls/atomic_structure_model_validator.gd
+++ b/godot_project/editor/controls/dockers/workspace_docker/simulations_docker/simulations_docker_controls/atomic_structure_model_validator.gd
@@ -182,10 +182,9 @@ func fix_overlapping_atoms() -> void:
 		for atom_id: int in current_overlap_atoms:
 			var item: SpatialHashGrid.Item = current_overlap_atoms[atom_id]
 			var position: Vector3 = item.position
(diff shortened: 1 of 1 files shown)


godot_project/editor/controls/dockers/workspace_docker/simulations_docker/simulations_docker_controls/atomic_structure_model_validator.gd
```diff
@@ -182,10 +182,9 @@ func fix_overlapping_atoms() -> void:
 		for atom_id: int in current_overlap_atoms:
 			var item: SpatialHashGrid.Item = current_overlap_atoms[atom_id]
 			var position: Vector3 = item.position
+			while position.is_equal_approx(overlap_center):
+				position = item.position + Vector3(randf() - 0.5, randf() - 0.5, randf() - 0.5)
 			var dir_from_center: Vector3 = overlap_center.direction_to(position)
-			while dir_from_center.is_equal_approx(Vector3.ZERO):
```
MarianoGnu, line 186: Turns with the more atoms where overlapping, the `sum(pos)/count` formula was accumulating enough error to make `overlap_center.direction_to(position)` return a value instead of `ZERO`, that also happened to be the same value for all atoms

HungryProton (approved):

PR #502, opened 2025-11-14
Forward modifier events to input handlers, even when mouse is on top of dynamic context panel

MarianoGnu:
- Note: issue only happened when hovering the dynamic context panel with the mouse while pressing the buttons, It also didn't replicate if Ctrl was slightly released before Option

Task: BUG - CNTL-Option can toggle app into a state where the editor does not match the 'Show Potential Atom Positions" control in the Workspace tab

### DIFF
--- a/godot_project/editor/controls/editor_viewport_container/editor_viewport.gd
+++ b/godot_project/editor/controls/editor_viewport_container/editor_viewport.gd
@@ -16,6 +16,10 @@ const INPUT_EVENT_ACTIONS_WHITELIST: Array[StringName] = [
 	&"toggle_ring_menu"
 ]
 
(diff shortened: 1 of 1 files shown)


HungryProton (changes requested):
https://github.com/user-attachments/assets/7b5c2276-831a-493c-ab18-ad36b88f1602

I don't remember the details, but I think this was related to the springs showing up when pressing shift above the dockers. This doesn't happen on devel, but it does in here

MarianoGnu:
ah, yes, basically not forwarding the events was on purpose to avoid this issue u.u
we can probably only forward ALT

HungryProton (approved):

PR #507, opened 2025-12-05
Added basic representation of DNA structures using path.

MarianoGnu:
- For now structures can only be created and destroyed with Undo/Redo and the snapshot system adds and removes the renderer

Task: Creating and Editing DNA


HungryProton (approved):

MarianoGnu:
why did this autoamtically merge?

PR #510, opened 2025-12-10
Transform Gizmo now works for DNA splines

MarianoGnu:
TasK: Creating and Editing DNA

https://github.com/user-attachments/assets/5c6e77c6-679e-457a-97ba-0e5dc61fdb77

https://github.com/user-attachments/assets/7877cbcf-6201-4b3b-97d1-bc372eca3041

### DIFF
--- a/godot_project/project_workspace/structs/dna_structure.gd
+++ b/godot_project/project_workspace/structs/dna_structure.gd
@@ -48,16 +48,27 @@ func grab_curve(out_path3d: Path3D) -> void:
 	out_path3d.curve = _curve
 
 
-func get_baked_path() -> PackedVector3Array:
-	if _baked_path.is_empty() or _signal_queue_path_changed:
-		_baked_path = _curve.get_baked_points()
+## Get a list of points representing the path
+## in_path_override can be provided to get an alternative path while editing the spline of this structure
+func get_baked_path(in_path_override: Curve3D = null) -> PackedVector3Array:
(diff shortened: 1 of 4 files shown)


godot_project/project_workspace/structs/dna_structure.gd
```diff
@@ -48,16 +48,23 @@ func grab_curve(out_path3d: Path3D) -> void:
...
 
-func get_baked_path() -> PackedVector3Array:
-	if _baked_path.is_empty() or _signal_queue_path_changed:
-		_baked_path = _curve.get_baked_points()
+## Get a list of points representing the path
+## in_path_override can be provided to get an alternative path while editing the spline of this structure
+func get_baked_path(in_path_override: Curve3D = null) -> PackedVector3Array:
+	var curve: Curve3D = _curve if in_path_override == null else in_path_override
+	if _baked_path.is_empty() or _signal_queue_path_changed or in_path_override != null:
+		var baked_path: PackedVector3Array = curve.get_baked_points()
```
HungryProton, line 60: /!\ get_baked_points() can return an empty array if the curve is empty (which would cause a crash on lines 59 - 60)

HungryProton (approved):

godot_project/project_workspace/structs/dna_structure.gd
MarianoGnu: i believe if all points where to be deleted, then the entire structure should be deleted, but having 1 point is likely an issue anyway

PR #513, opened 2025-12-12
Added basic representation of DNA structures using path.

MarianoGnu:
- For now structures can only be created and destroyed with Undo/Redo and the snapshot system adds and removes the renderer


MarianoGnu:
Previously approved as #507

PR #522, opened 2025-12-22
Add optional force limits to linear motors

HungryProton:
Card: Research how we might implement "force-based" motors.

---

[force_based_motors.webm](https://github.com/user-attachments/assets/c68b8999-2444-49bf-adcd-ec28663b83fe)

This only applies to Linear motors for now.

This works by extracting the force applied to the molecule to make it satisfy the motor constraint.

We have the previous velocity, and the new velocity imposed by the motor, using `Acceleration = Delta V / Time`, we compare the particle acceleration to the maximal theoretical acceleration (calculated from the user provided `max force`) and limit the new change in velocity if necessary.

~~Note: The UI mentions the force is in nanoNewtons, but that's not exactly true, it's in (Da . nm) / ns², and I'm not sure how to get a proper conversion to nN from there, if it even makes sense to begin with.~~

UI is now in attoNewtons (even 1fN is already too much in my test scene and working with small decimal numbers is not great from a UX perspective)

### DIFF
--- a/godot_project/python/scripts/openmm_server.py
+++ b/godot_project/python/scripts/openmm_server.py
@@ -409,6 +409,7 @@ class MotorForce:
 		self.motor_type: MotorType = motor_force_data["parameters"]["motor_type"]
 		
 		self.ramp_in_time_in_nanoseconds: float = motor_force_data["parameters"]["ramp_in_time_in_nanoseconds"]
+		self.limit_maximum_force: bool = motor_force_data["parameters"]["limit_maximum_force"]
 		match self.motor_type:
 			case MotorType.ROTARY:
 				rotary_polarity: RotaryPolarity = motor_force_data["parameters"]["polarity"]
@@ -425,6 +426,7 @@ class MotorForce:
 				linear_polarity: LinearPolarity = motor_force_data["parameters"]["polarity"]
 				self.polarity: float = 1.0 if linear_polarity == LinearPolarity.FORWARD else -1.0
 				self.top_speed_in_nanometers_by_nanoseconds = motor_force_data["parameters"]["top_speed_in_nanometers_by_nanoseconds"]
+				self.max_force = motor_force_data["parameters"]["max_force"] * 1e-18 # Convert attoNewtons to Newtons
 			case _:
 				logging.error(f"Unknown motor type '{self.motor_type}'!")
 				pass
@@ -537,7 +539,8 @@ class MotorForce:
 		self.cycle_started_to_stop_at_time: float = 0
 		self.cycle_pause_time_accum: float = 0.0
 		self.cycle_paused: bool = False
-		self.print_counter: int = 0
+		self.print_counter: int = -1
+		self.max_acceleration: float = -1.0
 		added_particles = 0
 		is_rotary: bool = self.motor_type == MotorType.ROTARY
 		for i in range(topology_payload.atoms_count):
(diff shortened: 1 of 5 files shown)


godot_project/python/scripts/openmm_server.py
```diff
@@ -646,6 +649,12 @@ def _advance_linear(self, simulation):
 			# Nothing to do here
 			return
 		
+		if not self.molecule_mass:
+			for i in range(len(self.atom_ids)):
+				atom_id: int = self.atom_ids[i]
+				self.molecule_mass = simulation.system.getParticleMass(atom_id)
```
MarianoGnu: ⛔  This doesn't look right, it is overriding `self.molecule_mass` with whatever is the last atom's isntead of acumulating it for calculating average
Also i would like check if atom_ids has only valid atoms form the molecule and doesn't include any kind of anchors used for locking atoms or for the Anchors&Sprints API. this is very unlikely, but is worth the check

godot_project/python/scripts/openmm_server.py
```diff
@@ -660,11 +669,22 @@ def _advance_linear(self, simulation):
...
+		max_acceleration: Vec3 = self.axis_direction * self.max_force / mass * (nanometer / nanosecond)
+		max_acceleration_len: float = length_Vec3(max_acceleration)
 		for i in range(len(self.atom_ids)):
 			atom_id: int = self.atom_ids[i]
-			current_velocity_in_desired_direction = self.axis_direction * dot_Vec3(prev_velocities[atom_id], self.axis_direction) * prev_velocities[atom_id].unit
-			delta_velocity = new_motor_velocity - current_velocity_in_desired_direction
-			new_velocities[atom_id] = prev_velocities[atom_id] + delta_velocity
+			previous_velocity: Vec3 = prev_velocities[atom_id]
+			final_velocity: Vec3 = new_motor_velocity
+			current_velocity_in_desired_direction = self.axis_direction * max(0.0, dot_Vec3(previous_velocity, self.axis_direction)) * previous_velocity.unit
```
MarianoGnu: any reason to clamp the velocity to zero? a negative velocity currently means hte particle is moving in oposite direction to the motor, what's the logic to ignore that?

godot_project/python/scripts/openmm_server.py
```diff
@@ -660,11 +669,22 @@ def _advance_linear(self, simulation):
 		state = simulation.context.getState(getVelocities=True)
 		prev_velocities = state.getVelocities()
 		new_velocities = prev_velocities.copy()
+		mass: float = self.average_particle_mass / dalton
```
MarianoGnu: nit: you could remove the unit once in line 656 and dont repeat a somewhat expenside operation every frame

MarianoGnu (changes requested):
Please check comments

godot_project/python/scripts/openmm_server.py
HungryProton: Done

godot_project/python/scripts/openmm_server.py
HungryProton: I read it wrong the first time and assumed this would amplify the motor speed if the particle was moving in the opposite direction. But that's not the case, I reverted it

godot_project/python/scripts/openmm_server.py
HungryProton: Good catch! 

I didn't do the check because the rest of the advance function doesn't check that either. I assume it should happen during the motor initialization instead but I'm not sure exactly

godot_project/python/scripts/openmm_server.py
```diff
@@ -646,6 +649,17 @@ def _advance_linear(self, simulation):
 			# Nothing to do here
 			return
 		
+		if self.limit_maximum_force and self.max_acceleration < 0:
+			average_particle_mass: float = 0.0
+			for i in range(len(self.atom_ids)):
+				atom_id: int = self.atom_ids[i]
+				average_particle_mass += simulation.system.getParticleMass(atom_id)._value
+			average_particle_mass /= len(self.atom_ids)
+			average_particle_mass *= 1.66053907e-27 # Convert Daltons to Kilograms
```
HungryProton, line 658: This conversion is done manually because openmm doesn't allow calling `dalton * kilogram` for some reason.

MarianoGnu (approved):
Cannot say about the math, but the code is correct

PR #524, opened 2025-12-23
Support converting a DNA Structure into groups of atoms

MarianoGnu:
Task. Creating and Editing DNA

### DIFF
--- a/godot_project/autoloads/dna_builder/DnaBuilder.gd
+++ b/godot_project/autoloads/dna_builder/DnaBuilder.gd
@@ -136,6 +136,10 @@ func get_template_atom_count(in_base: String, in_include_hydrogens: bool = false
 	return get_template(in_base, in_include_hydrogens).atoms.size()
 
 
+func get_template_bond_count(in_base: String, in_include_hydrogens: bool = false) -> int:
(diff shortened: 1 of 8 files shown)


HungryProton:
I'm not sure about this one. There's too much overlap with stuff that should only belong to AtomicStructure here (and I don't think making DNAStructure inherit AtomicStructure is a good idea either). 

Wouldn't it be a lot simpler if you created a new NanoMolecularStructure as a internal child of the DNA structure? (So the DNA structure only have to generate the atoms). 

That group doesn't have to be exposed to the user in the group list either, but this way you don't have to make a special renderer, or recreate the atom edition logic in two separate places.

MarianoGnu:
That would require allowing to use DnaStructure as a group, but with weird limitations, and that would require to capture special cases where you cannot edit an AtomicStructure only because is a child (only and special) or a DnaStructure....
After much tough this semmed to be the less invasive solution, but tbh i am not happy with anything u.u

HungryProton (approved):

HungryProton:
Okay, let's get it to a working state and when the feature / spec is working we can see if it's worth refactoring

PR #535, opened 2026-01-09
By default, 'Convert DNA template to atoms' will not override atom colors

MarianoGnu:
Task: In converting DNA to Atoms and Bonds, Override Colors Per Strand should default to Off


MarianoGnu:
Proactively merging since is trivial

PR #540, opened 2026-01-12
Define abstract class for MSEP Online Service

MarianoGnu:
(To be implemented as a REST client and as a Stub)

Task: SHARING FEATURE: Stub responses from rest service for testing in offline state.

### DIFF
--- a/godot_project/autoloads/msep_online/msep_online_service.gd
+++ b/godot_project/autoloads/msep_online/msep_online_service.gd
@@ -1,0 +1,395 @@
+"@abstract_class"
+class_name MsepOnlineService
+extends Node
+
+
+enum SortType { UPDATED, FIRST_CREATED, AUTHOR, TRENDING, POPULAR}
+
+
+class ServiceResponse:
+	var code := HTTPClient.ResponseCode.RESPONSE_NOT_FOUND
+	var body: Dictionary
+	var response_headers: PackedStringArray
+
+
+var authenticator: MsepOnlineAuthenticator:
+	get():
+		return MolecularEditorContext.authenticator
+
+
+@warning_ignore_start("unused_parameter")
+
+
+#region : Current User Endpoints
+## Get current authenticated user metadata
+func get_me() -> Promise:
+	assert(false, ClassUtils.ABSTRACT_FUNCTION_MSG)
+	return null
+
+
+## Update own profile
+func put_me(
+		username := OptionalString.empty(),
+		bio := OptionalString.empty(),
+		primary_email := OptionalString.empty(),
+		organization_affiliations := OptionalPackedStringArray.empty()
+	) -> Promise:
+	assert(false, ClassUtils.ABSTRACT_FUNCTION_MSG)
+	return null
+
+
+## List all projects the current user can edit
+func get_me_editable() -> Promise:
+	assert(false, ClassUtils.ABSTRACT_FUNCTION_MSG)
+	return null
+
+
+## Projects where the current user is listed as a collaborator
+func get_me_credited() -> Promise:
+	assert(false, ClassUtils.ABSTRACT_FUNCTION_MSG)
+	return null
+
+
+## List projects you've opted out of being credited on
+func get_me_opted_out() -> Promise:
+	assert(false, ClassUtils.ABSTRACT_FUNCTION_MSG)
+	return null
+
+
+## Delete own account
+func delete_me(confirmation: String, reason: String) -> Promise:
+	assert(false, ClassUtils.ABSTRACT_FUNCTION_MSG)
+	return null
+#endregion : Current User Endpoints
+
+
+#region : File Upload Endpoints
+## Request upload of user avatar
+func post_upload_user_avatar(file: FileAccess) -> Promise:
+	var file_path: String = file.get_path()
+	var extension: String = file_path.get_extension().to_lower()
+	if not _is_image_file(extension):
+		return _invalid_image_file_promise()
+	return _post_uploads("avatar", "user", "image/" + extension, file)
+
+
+## Request upload of project avatar
+func post_upload_project_avatar(in_namespace: String, project_name: String, file: FileAccess) -> Promise:
+	var file_path: String = file.get_path()
+	var extension: String = file_path.get_extension().to_lower()
+	if not _is_image_file(extension):
+		return _invalid_image_file_promise()
+	var additional_fields: Dictionary[String,String] = {
+		"namespace" : in_namespace,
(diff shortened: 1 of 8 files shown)


godot_project/autoloads/msep_online/msep_online_service.gd
```diff
@@ -0,0 +1,385 @@
...
+	assert(false, ClassUtils.ABSTRACT_FUNCTION_MSG)
+	return null
+
+
+## Update own profile
+func put_me(
+		username: Variant = null, # : String
+		bio: Variant = null, # : String
+		primary_email: Variant = null, # : String
+		organization_affiliations: Variant = null # : PackedStringArray
```
HungryProton: Rather than typing these as Variant so they can be nullable, it's probably better to give them their actual String type, with a default empty string / empty array.

Then when it's time to do the http request, if the string is empty, we don't include it.

I'm not sure the convienience of a null check is worth the increased error risk

godot_project/autoloads/msep_online/msep_online_service.gd
```diff
@@ -0,0 +1,385 @@
...
+
+
+## Request upload of project avatar
+func post_upload_project_avatar(in_namespace: String, project_name: String, file: FileAccess) -> Promise:
+	var file_path: String = file.get_path()
+	var extension: String = file_path.get_extension().to_lower()
+	if not extension in ["png", "jpg", "jpeg"]:
+		var fail := Promise.new()
+		fail.fail("Invalid image file format")
+		return fail
```
HungryProton: Could be worth to move in a helper method `is_image_file(file_access)` somewhere and reuse it above

godot_project/autoloads/msep_online/msep_online_service.gd
```diff
@@ -0,0 +1,385 @@
...
+	assert(false, ClassUtils.ABSTRACT_FUNCTION_MSG)
+	return null
+
+
+## Update profile (self or org admin)
+func put_namespace(
+		in_namespace: String,
+		bio: Variant = null, # : String
+		primary_email: Variant = null, # : String
+		organization_affiliations: Variant = null # : PackedStringArray
```
HungryProton: Same as l28, it's better to keep the actual types

godot_project/autoloads/msep_online/msep_online_service.gd
```diff
@@ -0,0 +1,385 @@
...
+	assert(false, ClassUtils.ABSTRACT_FUNCTION_MSG)
+	return null
+#endregion : Namespace Endpoints
+
+
+#region : Project Endpoints
+## List projects in a namespace
+func get_namespace_projects(
+		in_namespace: String,
+		sort:  String = "updated",
```
HungryProton: Since strings don't play nice with auto complete, we can use an enum instead and convert it into a string before the request.

The enum convertion can be a simple match with hardcoded strings, or it can be automated like this: 

```gdscript
enum SortType { UPDATED, FIRST_CREATED, AUTHOR }

func to_string(sort: SortType) -> String:
    return SortType.keys()[sort].to_lower()
```

godot_project/autoloads/msep_online/msep_online_service.gd
```diff
@@ -0,0 +1,385 @@
...
+
+
+## Update project metadata (including ownership transfer)
+func put_namespace_project(
+		in_namespace: String,
+		project_name: String,
+		description: Variant = null, # : String
+		collaborators: Variant = null, # : Array[Dictionary]
+		tags: Variant = null, # : PackedStringArray
+		transfer_to_new_owner: Variant = null, # String
```
HungryProton: Same as l28, better keep the types

godot_project/autoloads/msep_online/msep_online_service.gd
```diff
@@ -0,0 +1,385 @@
...
+		in_namespace: String,
+		project_name: String,
+		new_tags: PackedStringArray
+	) -> Promise:
+	assert(false, ClassUtils.ABSTRACT_FUNCTION_MSG)
+	return null
+
+
+## Add a single tag
+func post_namespace_project_tag(
```
HungryProton: I know these two functions were named after their api, but it may be better to rename them with more explicit names (like `replace_all_project_tags` and `add_project_tag`)

godot_project/autoloads/msep_online/msep_online_service.gd
```diff
@@ -0,0 +1,385 @@
...
+## List all tags with usage counts
+func get_explore_tags(sort: String = "popular", limit: int = 50) -> Promise:
+	assert(false, ClassUtils.ABSTRACT_FUNCTION_MSG)
+	return null
+
+
+## Projects with a specific tag
+func get_explore_tag_projects(
+		tag: String,
+		sort:  String = "updated",
```
HungryProton: same as l126

godot_project/autoloads/msep_online/msep_online_service.gd
```diff
@@ -0,0 +1,385 @@
...
+	return null
+#endregion : Version Permalink Endpoints
+
+
+#region : Discovery & Search Endpoints
+## Browse/search all public projects
+func get_explore_projects(
+		search: String,
+		tags: PackedStringArray,
+		sort: String = "trending",
```
HungryProton: same as l126

HungryProton (changes requested):

godot_project/autoloads/msep_online/msep_online_service.gd
MarianoGnu: What if someone actually wants their bio to be empty?
Particularly arguments of type PackedStringArray would be usefull to know if they should or not be used

godot_project/autoloads/msep_online/msep_online_service.gd
MarianoGnu: will name them `post_namespace_add_project_tag` and `put_namespace_replace_project_tags`

godot_project/autoloads/msep_online/msep_online_service.gd
HungryProton: I didn't really think of that, is the bio allowed to be empty?
Maybe we can just pass a whitespace or a special character in that case

godot_project/autoloads/msep_online/msep_online_service.gd
MarianoGnu: Will do, however i think different searches will use different sort tags, in example there's "popular" and "trending" in the examples, which are basically the same

HungryProton (approved):

PR #541, opened 2026-01-15
MSEP Online: Implement rest service endpoints, along with flags for using a stub service (in localhost) and run tests on it

MarianoGnu:
Task: SHARING FEATURE: Stub responses from rest service for testing in offline state.

https://github.com/user-attachments/assets/b4fbc739-17fd-46aa-80b9-f9cb2e69a106

[msep-online-test-logs.txt](https://github.com/user-attachments/files/24649158/msep-online-test-logs.txt)


godot_project/autoloads/msep_online/http/msep_online_http_request.gd
```diff
@@ -0,0 +1,94 @@
...
+			RESULT_TLS_HANDSHAKE_ERROR : "RESULT_TLS_HANDSHAKE_ERROR",
+			RESULT_NO_RESPONSE : "RESULT_NO_RESPONSE",
+			RESULT_BODY_SIZE_LIMIT_EXCEEDED : "RESULT_BODY_SIZE_LIMIT_EXCEEDED",
+			RESULT_BODY_DECOMPRESS_FAILED : "RESULT_BODY_DECOMPRESS_FAILED",
+			RESULT_REQUEST_FAILED : "RESULT_REQUEST_FAILED",
+			RESULT_DOWNLOAD_FILE_CANT_OPEN : "RESULT_DOWNLOAD_FILE_CANT_OPEN",
+			RESULT_DOWNLOAD_FILE_WRITE_ERROR : "RESULT_DOWNLOAD_FILE_WRITE_ERROR",
+			RESULT_REDIRECT_LIMIT_REACHED : "RESULT_REDIRECT_LIMIT_REACHED",
+			RESULT_TIMEOUT : "RESULT_TIMEOUT",
+		}
```
HungryProton, line 87: You don't have to redefine the whole thing manually, you can just convert the enum to string like this: 
`Result.keys()[result_code]`

godot_project/autoloads/msep_online/http/msep_online_upload_http_request.gd
```diff
@@ -0,0 +1,30 @@
...
+		"Content-Type: application/octet-stream",
+		"Content-Length: %d" % file.get_length()
+	]
+	var authenticator: MsepOnlineAuthenticator = MolecularEditorContext.authenticator
+	headers.append_array(authenticator.get_authentication_headers())
+	file.seek(0)
+	_promise.set_meta(&"url", "URL: "+ method + " : " + upload_url)
+	if _get_base_url().find("http://") == 0:
+		# Test server doesn't use a certificate
+		set_tls_options(TLSOptions.client_unsafe())
```
HungryProton: Maybe this should explicitly check if the test feature flag is enabled, so it only matters if we're connected to the stub server, instead of allowing any unsecured connections.

HungryProton (approved):

godot_project/autoloads/msep_online/http/msep_online_http_request.gd
MarianoGnu: i can only do that with enums defined in GdScript, this is a native enum, and is not defined as a Dictionary, so i cannot use `.keys()` on it (i tried)

image

godot_project/autoloads/msep_online/http/msep_online_upload_http_request.gd
MarianoGnu: tbh, if anyone ever overrides the api url (this is an opensource project after all) they will also have to provide valid certificates for it. I believe even we will have to once the real server is online. An https server should refuse the connection if a certificate is not provided anyway.

That being said, i have no clue of how tls protocol works and i may be flashing nonsense.

PR #559, opened 2026-02-02
DNA: Allow the user to specify a sequence length and autocomplete it with a random sequence

MarianoGnu:
Task: DNA: Allow to fill sequence with a random string instead of user specified, setting only the lenght of the chain in number of bases


kubecz3k (approved):
Style-wise looks good, don't see any obvious failure points, that being said haven't tried to understand the logic behind sequence methods

HungryProton (approved):

PR #615, opened 2026-02-26
Fix wrong emmision of selection change

MarianoGnu:
Task: BUG: Transform Gizmo not affecting atom(s) of a subgroup

### DIFF
--- a/godot_project/project_workspace/workspace_context/workspace_context.gd
+++ b/godot_project/project_workspace/workspace_context/workspace_context.gd
@@ -833,8 +833,8 @@ func _on_dna_path_changed(in_structure_context_id: int) -> void:
 			if not structure_context.has_selection():
 				# removed the last selection, let's keep selection alike
 				structure_context.select_all()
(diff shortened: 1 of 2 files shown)


godot_project/project_workspace/workspace_context/workspace_context.gd
```diff
@@ -833,8 +833,8 @@ func _on_dna_path_changed(in_structure_context_id: int) -> void:
 			if not structure_context.has_selection():
 				# removed the last selection, let's keep selection alike
 				structure_context.select_all()
-		var contexts: Array[StructureContext] = [structure_context]
-		selection_in_structures_changed.emit(contexts)
+			var contexts: Array[StructureContext] = [structure_context]
+			selection_in_structures_changed.emit(contexts)
```
MarianoGnu, line 837: This signal was being emitted when moving control points with the transform gizmo, which was not the intended case, and was affecting the position of moved objects, by resetting the transform gizmo in middle of `_apply_selection_transform()`

HungryProton (approved):

PR #631, opened 2026-03-04
Create UI to allow user select DNA Color overrides

MarianoGnu:
- UI is hidden for now
- Also made the Helix Properties panel collapsable

Task: DNA: Add settings to override atoms colors


godot_project/editor/controls/dockers/workspace_docker/shared_controls/dna_panel.gd
```diff
@@ -2,6 +2,8 @@ class_name  DnaPanel
 extends DynamicContextControl
 
 
+const COLOR_OVERRIDES_ENABLED = false
```
MarianoGnu, line 5: This is a temp flag, used instead of feature flags. Will delete it in the last PR, when the feature is fully functional

godot_project/editor/controls/dockers/workspace_docker/shared_controls/dna_panel.tscn
```diff
@@ -1,13 +1,79 @@
...
+func _on_pressed() -> void:
+	panel.visible = not panel.visible
+	_update_icon()
+
+
+func _update_icon() -> void:
+	const COLLAPSED_ICON: Texture2D = preload(\"uid://b2w2wxk2bb4wg\")
+	const EXPANDED_ICON: Texture2D = preload(\"uid://gl1hqwiod5y0\")
+	icon = EXPANDED_ICON if panel.visible else COLLAPSED_ICON
+"
```
HungryProton, line 38: Shouldn't that script be its own file instead of duplicating it multiple times everywhere?

HungryProton (approved):

godot_project/editor/controls/dockers/workspace_docker/shared_controls/dna_panel.tscn
MarianoGnu: because how specific it is i could not think in how could this be reused. If we encounter a case where script can be reused we can then save it to a separate file

PR #639, opened 2026-03-10
Set atoms colors in DNA Atoms&Bonds representation

MarianoGnu:
- Also enable the color properties in the create/edit dna panels

Task: DNA: Add settings to override atoms colors
image


godot_project/project_workspace/structs/dna_structure.gd
```diff
@@ -250,10 +251,38 @@ func end_edit() -> void:
...
+				if was_atom_added.call(atom_id) or was_atom_removed.call(atom_id):
+					return false
+				if prev_atoms_cache[atom_id].atomic_number != atom_get_atomic_number(atom_id):
+					_signal_queue_atomic_number_changed.append(Vector2i(atom_id, atom_get_atomic_number(atom_id)))
+					return true
+				return false
 			_signal_queue_atoms_removed = Array(all_old_atom_ids).filter(was_atom_removed)
 			_signal_queue_atoms_added = Array(all_new_atom_ids).filter(was_atom_added)
 			_signal_queue_atoms_moved = Array(all_new_atom_ids).filter(was_atom_moved)
+			Array(all_new_atom_ids).filter(check_atomic_number_changed)
```
HungryProton: Okay I see what it does but it took me a while because the previous filter pattern makes it look like this line (264) doesn't do anything. 

Maybe add a comment to say this is just a fast way to run a loop on the array and that's it's the lambda that's updating the `signal_queue_atomic_number_changed` directly

godot_project/project_workspace/structs/dna_structure.gd
```diff
@@ -250,10 +251,38 @@ func end_edit() -> void:
...
+			
+			var assign_atom_colors: Callable = func(atom_id: int) -> bool:
+				var expected_color_or_transparent: Color = _get_atom_expected_color(atom_id)
+				var should_have_color: bool = expected_color_or_transparent != Color.TRANSPARENT
+				if should_have_color:
+					set_color_override([atom_id], expected_color_or_transparent)
+				else:
+					remove_color_override([atom_id])
+				return false
+			Array(atoms_to_check_colors).filter(assign_atom_colors)
```
HungryProton: Same here

(Actually, this lambda doesn't even need the variables from the local block, so maybe it could be its own private method)

HungryProton (approved):

godot_project/project_workspace/structs/dna_structure.gd
MarianoGnu: i guess i can change it for a for loop, this was jsut more compart because takes 1 line instead of 2

HungryProton (approved):

PR #643, opened 2026-03-11
Fix a case where user could not move the camera and it should

MarianoGnu:
Task: BUG: can't move the camera after clicking on the group selector UI on top


godot_project/editor/controls/workspace_view/controls/structure_selector/structure_selector_bar.gd
```diff
@@ -179,6 +178,7 @@ func _on_structure_list_more_childs_requested(
 	list.toggle_what_to_show()
 	list.rebuild_list()
 	list.show()
+	_showing_more_childs = list.get_what_to_show() == StructureChildList.WhatToShow.ALL
```
MarianoGnu, line 181: Setting the right value to `_showing_more_childs` makes `_has_point` work as intended and not block mouse inputs

HungryProton (approved):

PR #651, opened 2026-03-13
Make sure DNA Object's atom colors are always tracked

MarianoGnu:
Task: BUG: DNA Object atom's color overrides are not carried over when copypastting an object

PR #652, opened 2026-03-13
Fix bonds renderer not picking color overrides from not NanoMolecularStructure's

MarianoGnu:
- This was affecting DNA Objects

Task: BUG: DNA bonds does not change color matching their atoms color override


MarianoGnu:
Note: moving the DNA Object and undoing the movement will make bonds look wrong again, this will be fixed when #651 is merged

HungryProton (approved):

PR #668, opened 2026-03-19
Prevent OpenMM from spaming prints when motors and particle emitters coexists in the same system

MarianoGnu:
Task: BUG: OpenMM print spam when motor and particle emitter in the same simulation


godot_project/python/scripts/openmm_server.py
```diff
@@ -936,7 +937,7 @@ def advance(self, simulation: Simulation):
 			self.enable_instance(i, simulation)
 
 	def enable_instance(self, instance_index: int, simulation: Simulation):
-		forces = SystemForcesCollection(simulation.system)
+		forces: SystemForcesCollection = self.forces
```
MarianoGnu, line 940: This is an optimization to avoid unnecesairly iterating all the forces every time a new instance is activated

HungryProton (approved):

PR #680, opened 2026-03-27
Fix crash when moving atoms with atom-to-atom springs to another group

MarianoGnu:
Task: CHASH: Moving atoms with atom-to-atom springs to a new group

### DIFF
--- a/godot_project/project_workspace/structs/atomic_structure.gd
+++ b/godot_project/project_workspace/structs/atomic_structure.gd
@@ -805,20 +805,30 @@ func merge_structure(in_structure: AtomicStructure, in_placement_xform: Transfor
 			new_bonds.push_back(bond_id)
 	
 	# Add Springs
+	var new_spring_id: int
 	for spring_id: int in in_structure.springs_get_valid():
 		if not in_structure.spring_has(spring_id):
 			continue
 		var related_anchor_id: int = in_structure.spring_get_anchor_id(spring_id)
 		var related_atom: int = in_structure.spring_get_atom_id(spring_id)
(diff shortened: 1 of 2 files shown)


godot_project/project_workspace/structs/atomic_structure.gd
```diff
@@ -805,20 +805,27 @@ func merge_structure(in_structure: AtomicStructure, in_placement_xform: Transfor
...
+		if related_anchor_id != Workspace.INVALID_OBJECT_INDEX:
+			var anchor: NanoVirtualAnchor = in_workspace.get_structure_by_int_guid(related_anchor_id)
+			if _check_if_anchor_connected_to_atom(anchor, new_atom):
+				continue
+			new_spring_id = spring_create(related_anchor_id, new_atom, constant_force,
+					equilibrium_length_is_auto, equilibrium_manual_length)
+		elif second_atom != AtomicStructure.INVALID_ATOM_ID:
+			new_spring_id = spring_create_between_atoms(new_atom, new_atom2, constant_force,
+					equilibrium_length_is_auto, equilibrium_manual_length)
 		new_springs.append(new_spring_id)
```
HungryProton, line 832: Probably wont happen, but if the second atom is invalid, the code skips the if / elif block and new_spring_id will be duplicated in new_springs

HungryProton (approved):

godot_project/project_workspace/structs/atomic_structure.gd
MarianoGnu: that's an assertable case, because spring has no target

PR #685, opened 2026-03-30
Add some filters to AutoCreateSprings panel to avoid creating springs that are unlikely to be desired

MarianoGnu:
- By default wont create springs to hydrogens
- (atom-to-atom springs) By default wont create springs linking atoms of the same molecule

Task: UI/UX: Add check buttons to ignore hydrogens when creating springs from selection

https://github.com/user-attachments/assets/e90a75e3-0089-45cc-9963-815f426fe872


godot_project/editor/controls/dockers/workspace_docker/shared_controls/toggle_button_controlled_visibility.tres.gd
```diff
@@ -0,0 +1,26 @@
+# This builtin script makes a control change it's visibility based on a toggle button
```
HungryProton: Comment is outdated, it's no longer a built-in script

Alos the ".tres" should probably be removed from the script file name

HungryProton (approved):

PR #703, opened 2026-04-07
Don't render bonds of length zero

MarianoGnu:
Task: BUG: Stray Bond


HungryProton (approved):

godot_project/editor/rendering/atomic_structure_renderer/representation/stick_representation/stick_representation.gd
```diff
@@ -713,6 +713,8 @@ func _calculate_partial_selection_translation(in_bond: Vector3i, in_delta_transl
 
 static func calculate_transform_for_bond(in_first_pos: Vector3, in_sec_pos: Vector3,
 			in_up_vector: Vector3) -> Transform3D:
+	if in_first_pos == in_sec_pos:
+		return Transform3D().scaled(Vector3.ZERO)
```
HungryProton, line 717: since we don't know why these 2 values could be equal, might be worth using `in_first_pos.is_equal_approx(in_sec_pos)` because of floating point precision

PR #709, opened 2026-04-08
Creating springs with shift + clic now respects the 'Ignore hydrogens` toggle

HungryProton:
Card: BUG - Ignore Selected Hydrogens no longer works when connecting springs to atoms

---

The `Ignore hydrogens` toggle from the `Auto create springs` panel works as expected.
But from what I understand from the card, I assume Jonathan tried to create springs with the shift+click shortcut, and thought the toggle didn't work.

This PR adds a new `spring_ignore_hydrogens` to the `create object parameters` class, and that setting is used when previewing and creating springs with shift+click. 

The UI was not changed at all, but I'm not sure where to move that `ignore hydrogen` toggle button.

### DIFF
--- a/godot_project/project_workspace/workspace_context/create_object_parameters.gd
+++ b/godot_project/project_workspace/workspace_context/create_object_parameters.gd
@@ -16,6 +16,7 @@ signal snap_to_shape_surface_changed(enabled: bool)
 signal spring_constant_force_changed(force: float)
 signal spring_equilibrium_length_is_auto_changed(is_auto: bool)
 signal spring_equilibrium_manual_length_changed(manual_length: float)
+signal spring_ignore_hydrogens_changed(enabled: bool)
 # validate_bonds_requested signal is used as intermediator between relax_tools_panel and validate_bonds_panel
 signal validate_bonds_requested(atom_set: AtomicStructure.AtomSet)
 
@@ -86,6 +87,8 @@ var _spring_constant_force: float = 5000000.0 # nN/nm
 var _spring_equilibrium_length_is_auto: bool = true
 
 var _spring_equilibrium_manual_length: float
+
+var _spring_ignore_hydrogens: bool = false
 
 var _default_shape: int = 0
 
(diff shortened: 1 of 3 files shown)


godot_project/editor/controls/dockers/workspace_docker/a_create_docker/auto_create_springs_panel.tscn
```diff
@@ -4,7 +4,7 @@
 [ext_resource type="PackedScene" uid="uid://b15453vqjum8" path="res://editor/controls/general/spin_box_slider.tscn" id="2_k222l"]
 [ext_resource type="PackedScene" uid="uid://b2b25o2443x3b" path="res://editor/controls/general/info_label.tscn" id="2_yoee2"]
 [ext_resource type="Texture2D" uid="uid://wbe2syrxllv8" path="res://editor/controls/menu_bar/menu_create/menu_atoms/icons/icon_AutoBonder_16px.svg" id="3_3tg0t"]
-[ext_resource type="Script" uid="uid://d2h7xyc1cxxeo" path="res://editor/controls/dockers/workspace_docker/shared_controls/toggle_button_controlled_visibility.gd" id="3_wjjra"]
```
MarianoGnu, line 7: this is fishy, why Godot editor tries to remove the uid? it happened the same to me a week ago or so, and i reverted this change

godot_project/project_workspace/workspace_context/create_object_parameters.gd
```diff
@@ -298,3 +301,12 @@ func set_spring_equilibrium_manual_length(in_manual_length: float) -> void:
 
 func get_spring_equilibrium_manual_length() -> float:
 	return _spring_equilibrium_manual_length
+
+
+func set_spring_ignore_hydrogens(in_enabled: bool) -> void:
+	_spring_ignore_hydrogens = in_enabled
+	spring_ignore_hydrogens.emit()
```
MarianoGnu: this isgnal is unused, if you want to keep it for consistency should be named `spring_ignore_hydrogens_changed`

MarianoGnu (commented):
Code is good.

I believe this is fundamentally wrong from the UX perspective, because a checkbutton from one panel is afecting a different one. I explicitly mentioned Jonathan this check button wouldn't affect shift+click when i made the feature in first place, but seems he wants this in

godot_project/project_workspace/workspace_context/create_object_parameters.gd
HungryProton: Signal has been renamed

HungryProton:
Yes, if Jonathan doesn't have a better idea, we could make a new panel right before "Creation distance" called 
"Spring Creation options" or something, and move that toggle there. But then it's not obvious that the toggle also affects shift click

Maybe it could be duplicated, and have two settings, one for the auto create, and the other for shift clic 

image

MarianoGnu:
Merge this and out the question in the card

PR #716, opened 2026-04-10
Prevent creating empty DNA

HungryProton:
Card: BUG: Possible to create empty DNA

---

Not sure what to do with the existing DNA when all bases are removed though.

### DIFF
--- a/godot_project/editor/controls/dockers/workspace_docker/a_create_docker/create_dna_panel.gd
+++ b/godot_project/editor/controls/dockers/workspace_docker/a_create_docker/create_dna_panel.gd
@@ -120,6 +120,11 @@ func _on_dna_sequence_text_edit_text_changed() -> void:
 	_update_create_button()
 
 
+func _sequence_button_button_group_pressed() -> void:
+	super._sequence_button_button_group_pressed()
+	_update_create_button.call_deferred()
+
+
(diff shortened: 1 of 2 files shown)


godot_project/editor/controls/dockers/workspace_docker/a_create_docker/create_dna_panel.gd
```diff
@@ -120,6 +120,11 @@ func _on_dna_sequence_text_edit_text_changed() -> void:
 	_update_create_button()
 
 
+func _sequence_button_button_group_pressed() -> void:
+	super._sequence_button_button_group_pressed()
+	_update_create_button.call_deferred()
```
MarianoGnu, line 125: Why was call deferred needed?

MarianoGnu (approved):

godot_project/editor/controls/dockers/workspace_docker/a_create_docker/create_dna_panel.gd
HungryProton: because this is called once during the NOTIFICATION_INSTANTIATED from the parent class, before the child class notification. So at that point, create_button is not initialized yet and I didn't wanted to duplicate the initialization in the update function.

PR #723, opened 2026-04-14
Fix the snap offset when using the plane transform gizmo

HungryProton:
Card: BUG: Shift to snap axis is broken when using the ✥ transform handle


MarianoGnu:
Note to self: Ortho surface X and Y probably would have the same issue, but we are safe because they are unused

MarianoGnu (approved):

PR #740, opened 2026-04-21
Fix crash applying atoms to shapes with collision detection

MarianoGnu:
Task: CRASH - Filling overlapping shapes while using the Don't Create Atoms in the Occupied space feature


MarianoGnu:
Merged on Jonathan's request

PR #789, opened 2026-05-27
Pick a default face when chosing 'Specific Box Plane'

HungryProton:
Card: BUG - When selecting Align to Specific Box Plain, the program should arbitrarily select a face from the currently selected Object


godot_project/editor/controls/dockers/workspace_docker/c_dynamic_context_docker/dynamic_context_controls/align_selection_panel.gd
```diff
@@ -153,6 +153,8 @@ func _on_relative_to_option_button_item_selected(in_index: int) -> void:
 	_align_selection_parameters.set_align_relative_to(
 		in_index as AlignRelativeTo
 	)
+	if (in_index as AlignRelativeTo) == AlignRelativeTo.SPECIFIC_BOX_PLANE:
+		_advance_specific_box_face(0)
```
MarianoGnu, line 157: This change is not compatible with https://github.com/MSEP-one/msep.one/pull/790

I am solving this issue in a kinda different way, each box has it's own selected_face and align_to_face, and in the case of align_to_face can never be "undefined"

PR #795, opened 2026-06-01
Attempt to generate more consistent Oriented Bounding Boxes from circle-shaped atom groups

MarianoGnu:
- NOTE: This will not generate the expected box from a Plane covered with atoms. However the result seems to be better than the one created before

Task: UX: Atoms arranged in a circle shape generates an OBB with very arvitrary rotation

|Before|After|
|--|--|
|image|image|
|image|image|


HungryProton:
I'm not convinced forcing the world Y on every square OBB will fix the issue without causing a lot of other problems.

I'll try the cache approach I was thinking about (keep the same obb as long as the individual atoms aren't moved).

MarianoGnu:
superseeded by #805

PR #800, opened 2026-06-02
Fix infinite loop caused by malformed obb

HungryProton:
Card: BUG: selecting exactly two atoms causes an infinite loop when the align panel is open

---

`AABB.has_surface()` only checks if at least one of the three size component is greater than 0. But if there's only one, it's a line, not a face. Not sure if it's worth notifying upstream.


MarianoGnu (approved):
I'd definitively report this

PR #810, opened 2026-06-04
Minimize volume of Alignable boxes

MarianoGnu:
Task: BUG - Algorithm does not correctly find faces for cube object

### DIFF
--- a/godot_project/project_workspace/workspace_context/structure_context/structure_context.gd
+++ b/godot_project/project_workspace/workspace_context/structure_context/structure_context.gd
@@ -470,6 +470,23 @@ func get_selection_aabb() -> AABB:
 	return _selection_db.get_selection_aabb()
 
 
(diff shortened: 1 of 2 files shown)


godot_project/project_workspace/workspace_context/structure_context/structure_context.gd
```diff
@@ -470,6 +470,23 @@ func get_selection_aabb() -> AABB:
 	return _selection_db.get_selection_aabb()
 
 
+func count_obb_with_selection_policy(in_policy: AlignSelectionGroupingPolicy) -> int:
```
MarianoGnu, line 473: The purpose of adding this method is to not require a potentially expensive calculation of the boxes sizes only to know if the panel should be shown

HungryProton (approved):

PR #816, opened 2026-06-11
Fix AlignSelectionPanel aligning rotation of only one of the selected groups

MarianoGnu:
Task: BUG - When multiple objects to align are selected, pressing the align button only aligns one of them


MarianoGnu:
Merged at Jonathan's request

PR #817, opened 2026-06-11
Fix align rotation math

MarianoGnu:
(for like the 5th time)

Task: BUG - Motor Alignment no longer works


MarianoGnu:
MErged at Jonathan's request

PR #829, opened 2026-06-23
Create carbon nanotube when releasing mouse button on create nanotube mode

MarianoGnu:
Task: Single Wall Carbon Nano-Tube Creator

https://github.com/user-attachments/assets/3f2f8096-7fdd-459d-bf8d-34040e9c031b

PR #831, opened 2026-06-26
Insert AtomicVirtualStructure class in the inheritance chain of DnaStructure

MarianoGnu:
- Next: Create CarbonNanotubeStructure inheriting AtomicVirtualStructure

Task: Single Wall Carbon Nano-Tube Creator

**This is a reopening of #830 which was accidentally merged into #829 instead of devel**


MarianoGnu:
This PR is preapproved in #830

PR #843, opened 2026-07-06
Center the alignment planes on the box

HungryProton:
Card: Alignment planes should appear once (no longer like boxes), and should be placed passing through the center of mass

---

Changes:

+ Keep 3 alignment planes (instead of all 6 faces of the box, opposite faces are removed)
+ The active alignment plane follows the align depth / center of mass properties
image
 
+ The advanced depth slider is now a CustomSpinBox for feature parity with the horizontal and vertical sliders
image

### DIFF
--- a/godot_project/editor/controls/dockers/workspace_docker/c_dynamic_context_docker/dynamic_context_controls/align_selection_panel.gd
+++ b/godot_project/editor/controls/dockers/workspace_docker/c_dynamic_context_docker/dynamic_context_controls/align_selection_panel.gd
@@ -244,12 +244,9 @@ func _on_alignable_boxes_tree_button_clicked(item: TreeItem, _column: int, id: i
 func _get_box_face_icon(selected_face: BoxFace) -> Texture2D:
 	const ICONS: Dictionary[BoxFace, Texture2D] = {
 		BoxFace.UNDEFINED : preload("uid://dr6k8q285dgls"),
-		BoxFace.FRONT  : preload("uid://d16yxekfb4k3h"),
-		BoxFace.BACK   : preload("uid://0vpkfnqii60i"),
-		BoxFace.TOP    : preload("uid://de1hnqihu4ugw"),
-		BoxFace.BOTTOM : preload("uid://b8qwlg7asmnsf"),
-		BoxFace.LEFT   : preload("uid://deu8o23ue70ft"),
-		BoxFace.RIGHT  : preload("uid://c01w40g4dq4eg"),
+		BoxFace.XY : preload("uid://d16yxekfb4k3h"),
+		BoxFace.XZ : preload("uid://de1hnqihu4ugw"),
+		BoxFace.YZ : preload("uid://c01w40g4dq4eg"),
 	}
 	return ICONS[selected_face]
 
--- a/godot_project/editor/rendering/align_selection_preview/align_selection_preview.gd
+++ b/godot_project/editor/rendering/align_selection_preview/align_selection_preview.gd
@@ -86,15 +86,6 @@ func _process(_delta: float) -> void:
 
 
 func _draw() -> void:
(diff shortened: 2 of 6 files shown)


godot_project/editor/controls/dockers/workspace_docker/c_dynamic_context_docker/dynamic_context_controls/align_selection_panel.gd
```diff
@@ -244,12 +244,9 @@ func _on_alignable_boxes_tree_button_clicked(item: TreeItem, _column: int, id: i
...
 		BoxFace.UNDEFINED : preload("uid://dr6k8q285dgls"),
-		BoxFace.FRONT  : preload("uid://d16yxekfb4k3h"),
-		BoxFace.BACK   : preload("uid://0vpkfnqii60i"),
-		BoxFace.TOP    : preload("uid://de1hnqihu4ugw"),
-		BoxFace.BOTTOM : preload("uid://b8qwlg7asmnsf"),
-		BoxFace.LEFT   : preload("uid://deu8o23ue70ft"),
-		BoxFace.RIGHT  : preload("uid://c01w40g4dq4eg"),
+		BoxFace.XY : preload("uid://d16yxekfb4k3h"),
+		BoxFace.XZ : preload("uid://de1hnqihu4ugw"),
+		BoxFace.YZ : preload("uid://c01w40g4dq4eg"),
```
MarianoGnu, line 249: Note, i didn't like the XY, XZ, YZ naming schema because it takes a bigger cognitive load to understand what name corresponds to what plane of the box

HungryProton: I used these to match what's being done in `AlignSelectionParameters.WorldPlane` but I can revert to FRONT / TOP / SIDE if you prefer, but these imply a direction

godot_project/editor/rendering/align_selection_preview/align_selection_preview.gd
```diff
@@ -195,9 +168,6 @@ func _draw_obb_reference_point(in_obb: AlignableOBB, in_face: BoxFace, in_highli
 			no_depth_ref_point = ref_point
 			if _align_selection_parameters.is_align_depth_enabled():
 				ref_point += basis.z * (in_obb.offset_ratio_d * face_size.z)
-			else:
-				ref_point += basis.z * (face_size.z * 0.5)
-
```
MarianoGnu, line 200: i am not sure if this offset should be removed when align depth is disabled, because depth still could affect the math when boxes are unaligned

MarianoGnu (approved):

godot_project/editor/rendering/align_selection_preview/align_selection_preview.gd
HungryProton: That one is just a visual change, if we keep the offset, the reference point is placed on the side of the OBB instead of in the center. 

It doesn't affect the positioning logic

Image

MarianoGnu: It might as well affect positioning, see how changing depth affects the anchoring

https://github.com/user-attachments/assets/b2deaad0-538a-44e6-aa00-b118c99b0f53

This same displacement should affect positioning when align depth is disabled

HungryProton: But align depth is enabled on that example, and if it's disabled, then it only moves the objects to be on the same axis.

I tried a lot of different setups, I don't see a single case where it would cause an issue

MarianoGnu: yes, the proposal is to allow change depth even when align depth is disabled

PR #845, opened 2026-07-06
Select the squarest face by default when opening the align panel

HungryProton:
Card: When alignment planes are set, the 1st plane selected should be most likely the user will want to align to.

### DIFF
--- a/godot_project/project_workspace/workspace_context/alignable_oriented_bounding_box.gd
+++ b/godot_project/project_workspace/workspace_context/alignable_oriented_bounding_box.gd
@@ -49,6 +49,8 @@ func _init(
 	_params = in_align_parameters
 	if not has_face(align_to_face):
 		advance_align_to_face(1)
+	align_to_face = _get_squarest_face()
+	selected_face = align_to_face
 	_initialized = true
 
 
@@ -327,3 +329,22 @@ func _get_align_reference_point(reference_face: BoxFace, in_align_depth: bool) -
 		reference_point += basis.z * size.z * DEFAULT_DEPTH_RATIO
 	
 	return transform.origin + reference_point
+
+
+## Returns the most square face (aspect ratio closest to 1.0) as it's the most
+## likely face to align a rotary motor with.
+## Returns BoxFace.UNDEFINED with the OBB has no valid surfaces.
+func _get_squarest_face() -> BoxFace:
+	var face: BoxFace = BoxFace.UNDEFINED
+	var squarest_ratio: float = INF
+	for idx in range(3):
+		var face_size: Vector3 = get_face_size(idx)
+		var smallest: float = min(face_size.x, face_size.y)
+		var largest: float = max(face_size.x, face_size.y)
+		if is_zero_approx(smallest):
+			continue # Skip division by zero
+		var ratio: float = largest / smallest
+		if face == BoxFace.UNDEFINED or ratio < squarest_ratio:
+			face = idx as BoxFace
+			squarest_ratio = ratio
+	return face


MarianoGnu:
Right now the front face is selected to be the biggest face, maybe instead of making the biggest face the front and autoselect, make the squarest face to be the front face.

Opinions?

godot_project/project_workspace/workspace_context/alignable_oriented_bounding_box.gd
```diff
@@ -327,3 +328,21 @@ func _get_align_reference_point(reference_face: BoxFace, in_align_depth: bool) -
...
+## Returns the most square face (aspect ratio closest to 1.0) as it's the most
+## likely face to align a rotary motor with.
+func _get_squarest_face() -> BoxFace:
+	var face: int = -1
+	var threshold: float = -1.0
+	for idx in range(3):
+		var face_size: Vector3 = get_face_size(idx)
+		if is_zero_approx(face_size.y):
+			continue # Skip division by zero
+		var ratio: float = abs((face_size.x / face_size.y) - 1.0)
```
MarianoGnu: i'm not understanding the involved math here, why substract 1, and why default thresshold is -1?
ratio is an abs(something) making impossible to `ratio < threshold`to be true, so this function will simply pick the first face with non size.y (only because face is -1)

MarianoGnu: i think it should go something like this, but i didn't test it:
```gdscript
## Returns the most square face (aspect ratio closest to 1.0) as it's the most
## likely face to align a rotary motor with.
func _get_squarest_face() -> BoxFace:
	var face: BoxFace= BoxFace.UNDEFINED
	var squarest_ratio: float = INF
	for idx in range(3):
		var face_size: Vector3 = get_face_size(idx)
		if is_zero_approx(face_size.y):
			continue # Skip division by zero
		var ratio: float = abs(face_size.x / face_size.y)
		if face == BoxFace.UNDEFINED or abs(ratio - 1.0) < squarest_ratio:
			face = idx as BoxFace
			squarest_ratio= abs(ratio - 1.0)
	return face
```

MarianoGnu (changes requested):

godot_project/project_workspace/workspace_context/alignable_oriented_bounding_box.gd
HungryProton: Default threshold doesn't matter, it's ignored on the first loop iteration because the face is undefined.

A perfect square has a ratio of `x / y = 1`.  if y is larger than x, then the ratio gets closer to 0.0 so just checking the smallest one won't work, it has to the closest to 1.0.

So the -1.0 part is just to bring it back to zero, and abs() to prevent negative results that would be picked in priority. Then the smallest result becomes the closest to 0.
I realize my variable names aren't accurate though, I'll fix it.

HungryProton: And it might not work properly with ratios between 0-1 actually

HungryProton: I've changed it so:
+ align_to_face and selected_face are the same on init
+ the ratio is always calculated as largest / smallest

the issue with the abs and -1.0 is that it always favor the tall narrow faces over everything else. (since these ones are stuck in the [0, 1] range and the others are in the [1, INF] range

godot_project/project_workspace/workspace_context/alignable_oriented_bounding_box.gd
```diff
@@ -327,3 +329,23 @@ func _get_align_reference_point(reference_face: BoxFace, in_align_depth: bool) -
...
+		var smallest: float = min(face_size.x, face_size.y)
+		var largest: float = max(face_size.x, face_size.y)
+		if is_zero_approx(smallest):
+			continue # Skip division by zero
+		var ratio: float = largest / smallest
+		if face == BoxFace.UNDEFINED or ratio < squarest_ratio:
+			face = idx as BoxFace
+			squarest_ratio = ratio
+	if face == BoxFace.UNDEFINED: # OBB has no valid face
+		return BoxFace.UNDEFINED
```
MarianoGnu: This is the same as `return face`

godot_project/project_workspace/workspace_context/alignable_oriented_bounding_box.gd
```diff
@@ -327,3 +329,23 @@ func _get_align_reference_point(reference_face: BoxFace, in_align_depth: bool) -
...
+	return face as BoxFace
```
MarianoGnu: face is already BoxFace

MarianoGnu (approved):

PR #875, opened 2026-07-16
Fix graphene lattice preview picking the wrong viewport texture for post-procesing when Alerts window is visible

MarianoGnu:
BackBufferCopy node ensures the screen texture is the expected one

Task: BUG - Nanotube Graph becomes transparent when "Alert" box is up


HungryProton:
If you're okay with a slight visual change, I'd rather not use the screen texture at all (see #876) rather than adding a BackBufferCopy as these are awfully expensive on lower end hardware 
image

MarianoGnu:
Closed in favor of #876

PR #878, opened 2026-07-20
Avoid double signal connection

MarianoGnu:
Task: BUG: Error spam when undog convert nanotube to atoms and bonds

### DIFF
--- a/godot_project/editor/controls/dockers/workspace_docker/c_dynamic_context_docker/dynamic_context_controls/edit_carbon_nanotube.gd
+++ b/godot_project/editor/controls/dockers/workspace_docker/c_dynamic_context_docker/dynamic_context_controls/edit_carbon_nanotube.gd
@@ -70,7 +70,8 @@ func _set_selected_context(out_nanotube_or_null: CarbonNanotubeStructure) -> voi
 			_edited_nanotube.is_trim_invalid_valence_carbons_enabled()
 		)
 		_length_spin_box_slider.set_value_no_signal(_edited_nanotube.get_tube_length())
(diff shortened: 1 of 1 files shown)


godot_project/editor/controls/dockers/workspace_docker/c_dynamic_context_docker/dynamic_context_controls/edit_carbon_nanotube.gd
```diff
@@ -70,7 +70,8 @@ func _set_selected_context(out_nanotube_or_null: CarbonNanotubeStructure) -> voi
 			_edited_nanotube.is_trim_invalid_valence_carbons_enabled()
 		)
 		_length_spin_box_slider.set_value_no_signal(_edited_nanotube.get_tube_length())
-		_edited_nanotube.path_changed.connect(_on_edited_nanotube_path_changed.unbind(2))
+		if not _edited_nanotube.path_changed.is_connected(_on_edited_nanotube_path_changed):
+			_edited_nanotube.path_changed.connect(_on_edited_nanotube_path_changed.unbind(2))
```
MarianoGnu, line 74: Logic thinking tells me this check is not needed, but the error is still happening, is like the disconnect in line 62 is not happening. [I wonder if this is related to signal connection snapshots that Jakub did?](https://github.com/MSEP-one/msep.one/blob/devel/godot_project/project_workspace/structs/atomic_structure.gd#L1054) even if that was the case the signal would not be recreated in the same frame, so was signal fully duplicated?

HungryProton (approved):

PR #947, opened 2026-08-26
Record Simulation Video: Ensure timestamp label is visible on any background color possible

MarianoGnu:
Task: BUG: Timestamp label needs contrast with bright backgrounds

**Note:** This PR makes timestamp label very big, this is because outline size is for some reason limited to 1/3 of the font size, and in order to guarrantee outline was big enough to produce contrast an arbitrary size of 48 was chosen

image


HungryProton (changes requested):
That means the font size controls no longer works (but they're still here)
Instead of a fixed size, what do you think of a solid background that the user can turn on and off ? #950

PR #950, opened 2026-08-27
Add an optional background to the videos timestamp

HungryProton:
Card: BUG: Timestamp label needs contrast with bright backgrounds

This PR adds an optional solid background to the timestamp text.
(color could be made user selectable in the future if needed)

Superseeds #947 

image

image


MarianoGnu (changes requested):
Label has a stylebox override that would be much easier to use for a background than all this Nodes hierarchy

image

HungryProton:
I never noticed that property before :+1: 
I removed the extra panel & margin container.

MarianoGnu (approved):

PR #961, opened 2026-09-21
Revert the workspace to its previous valid state when aborting a simulation

HungryProton:
Card: After Cancelling a simulation, anchors not selected


godot_project/project_workspace/workspace_context/history/history.gd
```diff
@@ -199,6 +199,12 @@ func apply_next_snapshot() -> void:
 	next_snapshot_applied.emit(snapshot_name)
 
 
+func apply_current_snapshot() -> void:
+	_apply_snapshot_from_stack(_stack_pointer)
+	changed.emit()
```
MarianoGnu, line 204: History havent really changed, is this signal required for the fix to work?

MarianoGnu (approved):
Code makes sense. I haven't tested

godot_project/project_workspace/workspace_context/history/history.gd
HungryProton: Going by the comment on the signal declaration, I assume it should trigger, and in practice, changed is always triggered at the same time as snapshot_applied. But these too are kinda redundant so maybe it could work without, but in doubt I left it there.

